Issue #21, opened 2023-08-08
Use available port if default port is already bound


srid:
```
❯ nix run github:juspay/nix-browser
Launching http://127.0.0.1:3000
```

This works, but if the user already has a process bound to port 3000 this will fail. We want nix-browser to automatically find an available port and use that if the default port is unavailable.

Issue #13, opened 2023-08-02
Add CLI args, and `--no-open`


srid:
- Use the `clap` crate to parse command line args into a Rust struct; [see example](https://github.com/srid/rust-nix-template/blob/8d51c03a1d52b8057724612246ba3776184f8723/src/main.rs#L3C1-L17), or this [more complex example](https://github.com/srid/nixci/blob/master/src/cli.rs)
- Use the [`open`](https://docs.rs/open/latest/open/) crate to automatically open the app in the user's browser, unless `--no-open` is passed (which we'll use for #11)
    - ~~Both~~ `nix run` ~~the devshell's `just watch`~~ should open in the default browser

srid:
This feature will be perfected by #21 - because a mere `nix run github:juspay/nix-browser` should immediately open the app in the user's browser without error'ing out.

PR #24, opened 2023-08-08
e2e testing using `thirtyfour`

### LINKED ISSUE #11: Add end to end testing
We want to test the final `nix build` web app. 

The official leptos templates use [playwright](https://playwright.dev/) to do this:

https://github.com/leptos-rs/start-axum/tree/main/end2end

While we may do the same, I'd like to first explore (briefly) the option of being able to write tests in pure Rust.  For eg., there is playwright-rust, but it seems [unmaintained](https://github.com/octaltree/playwright-rust/pull/49#issuecomment-1564119957). Are there other options? 

Exhausting all that, we may just use something like playwright.

divinenaman:
fixes #11


flake.nix
```diff
@@ -48,6 +48,7 @@
           nativeBuildInputs = with pkgs; [
             just
           ];
+          # packages = with pkgs; [ chromium ];
```
srid, line 51:
```suggestion
          packages = with pkgs; [
            just
          ] ++ lib.optionals pkgs.stdenv.isLinux [
            chromium
          ];
```

Cargo.toml
```diff
@@ -29,6 +29,7 @@ tokio = { version = "1.29", features = ["full"], optional = true }
 tower-http = { version = "0.4", features = ["full"], optional = true }
 tower = { version = "0.4", optional = true }
 http = { version = "0.2", optional = true }
+thirtyfour = "0.31.0"
```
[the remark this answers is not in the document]
srid: Also, shouldn't this be a dev-only dependency? cf. https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html

justfile
```diff
@@ -10,7 +10,15 @@ watch *ARGS:
...
-test:
+chromedriver:
+    #!/bin/sh
+    chromedriver &
+    echo $! > .chrome.pid
+
+kill-chromedriver:
+    kill -9 $(cat .chrome.pid)
+
+test: chromedriver
```
srid, line 21: This should be nixified. @shivaraj-bh Could you pair with @divinenaman on setting up process-compose based tests run as part of flake checks?

srid: These tests should, ultimately, be run in CI. Presently, in this PR, they do not.

divinenaman: @srid @shivaraj-bh I have added process-compose, do review it. I have question about how to integrate the same in CI, do I write it in jenkins itself ?

srid: I'm unable to test your branch. `just test` hangs at this stage:

image

---

I don't think running `cargo leptos watch` in process-compose is a good idea. Let's run `cargo leptos watch` by hand in one terminal, and your test runner in other terminal. 

Once I did this, the tests ran ... then failed:

image



---

As for CI, you basically want to use the Nix package directly:

```diff
diff --git a/flake.nix b/flake.nix
index 87da3b8..0c1c6a6 100644
--- a/flake.nix
+++ b/flake.nix
@@ -85,7 +85,7 @@
                   };
                 };
                 start-app = {
-                  command = "cargo leptos watch";
+                  command = self'.packages.default;
                   readiness_probe = {
                     exec.command = with pkgs; "${curl}/bin/curl --fail localhost:3000";
                     initial_delay_seconds = 2;
```

And then add a 'test' process that checks the result (@shivaraj-bh can help here).

But we also want a CLI option to pass `--port`, because the Nix package will use a random. We're waiting for @1vipulgupta 's PR for #13 before doing that.

srid:
@divinenaman By the way, you don't have to maintain a clean commit history and do rebases (I'm going to squash merge this PR as a single commit anyway.). Just `git merge` is fine; it should make conflict resolution simpler.

justfile
srid: You just have to change your tests, I don't see how you can automate it. 

Once the PR is merged, we won't be pushing commits that break tests.

srid: The other thing you can do is, don't merge from `main` until before this PR is ready for review/merge.

divinenaman: @srid passing CLI arguments to test will be bit awkward, cargo test internally handles arguments after `--`. To support CLI args, we need to create a `main` function that will execute all test. [Stack overflow Thread](https://stackoverflow.com/questions/65791697/passing-custom-command-line-arguments-to-a-rust-test). 

Apart from this, how will we know the port on which the server is running after #21, as I discussed with @1vipulgupta  about passing port 0 to axum, which will bind an available port ?

srid: > @srid passing CLI arguments to test will be bit awkward, cargo test internally handles arguments after `--`. To support CLI args, we need to create a `main` function that will execute all test. [Stack overflow Thread](https://stackoverflow.com/questions/65791697/passing-custom-command-line-arguments-to-a-rust-test).

The CLI option is for main program (nix-browser), not the tests. I've opened #43 to define it precisely.

> Apart from this, how will we know the port on which the server is running after #21, as I discussed with @1vipulgupta about passing port 0 to axum, which will bind an available port ?

We should have process-compose use a specific port for nix-browser via #43, and then the test can simply use the same port.

divinenaman:
@srid finalized changes. Need help with this error:

error:
qq

cause:
`command = "${self'.packages.default}/bin/nix-browser --site-addr 127.0.0.1:${env.TEST_PORT}";`

srid:
@divinenaman There was a test regression in `main` today, it got fixed in 95e4ea2925270ed186230c11f417fff2a9a0537e. Try latest `main`.

Once the PR is finalized, let's get on a call to go over it. When you are ready, book a time here: https://cal.com/sridca

srid:
Hmm, even after a merge from `main`, `nix build -L` fails on your branch for some reason:

image

srid:
Pushed a fix for that, now I see:

```
nix-browser> failures:
nix-browser> ---- test_nix_version stdout ----
nix-browser> Error: NewSessionError(Failed(hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message:
"Connection refused" }))))
nix-browser> failures:
nix-browser>     test_nix_version
nix-browser> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

---

I just realized that `test_nix_version` is actually coming from this PR (not `main` branch tests). Why is the nix build running e2e tests? It should not be running them; nix builds are pure, and you can't access the network.

divinenaman:
@srid please review, bugs fixed, not clear why CI is failing, local everything seems to work

flake.nix
```diff
@@ -25,56 +25,59 @@
...
       ];
-      perSystem = { config, self', pkgs, lib, system, ... }: {
-        _module.args.pkgs = import inputs.nixpkgs {
-          inherit system;
-          overlays = [
-            inputs.rust-overlay.overlays.default
-          ];
-        };
+      perSystem = { config, self', pkgs, lib, system, ... }:
+        {
```
srid, line 31: If you move this `{` to the line above (which is how it is in `main`), it will fix the indentation change that has been making the diff harder to read.

divinenaman: I see

divinenaman:
@srid playwright test also works on darwin

srid:
@divinenaman 

https://github.com/juspay/nix-browser/issues/11#issuecomment-1704227153

> @srid playwright test also works on darwin

That definitely tips the scale towards going with playwright. 

I've confirmed that `nix run .#cargo-e2e-playwright-test` runs successfully on my macbook, but there are a few issues:

1. It runs indefinitely, rather than exiting immediately after the tests finish running (or when one of them fail).
2. It seems to use the local Chromium browser (Brave in my case), as it immediately opens Brave in a new tab. But we want to run e2e in headless mode in CI.

Put simply, can we run e2e tests in Github Actions?

srid:
> 2\. It seems to use the local Chromium browser (Brave in my case), as it immediately opens Brave in a new tab. But we want to run e2e in headless mode in CI.

Ah, regarding this. I just noticed that it runs successfully on a headless Linux (ssh'ed to a remote server) as well. So we just want `nix run .#cargo-e2e-playwright-test` to exit immediately with the appropriate exit code depending on test result. So (1) needs to be done. I can see the indefinitely-run behaviour with the report server being desirable in development mode, though.

e2e-playwright/package.json
srid, line 1: Why do we need `package.json` and `package-lock.json`? The tests run fine without these.

e2e-playwright/flake-module.nix
```diff
@@ -0,0 +1,40 @@
...
+      process-compose.cargo-e2e-playwright-test =
+        let
+          TEST_PORT = "5000";
+        in
+        {
+          tui = false;
+          port = 8975;
+          settings.processes = {
+            start-app = {
+              command = "${lib.getExe self'.packages.default} --site-addr=127.0.0.1:${TEST_PORT}";
```
srid, line 15:
```suggestion
              command = "${lib.getExe self'.packages.default} --no-open --site-addr=127.0.0.1:${TEST_PORT}";
```

srid:
Regarding conflict resolution issues (why CI fails right now), maybe just start a new branch/PR containing _only_ playwright tests (without thirtyfour stuff, thus without any Cargo.lock changes). It is likely that we are going with playwright anyway.

srid:
Let's go with #54 !

PR #26, opened 2023-08-09
using available port if default is bound

1vipulgupta:
Used `try_bind` instead of `bind` which will panic if port is busy, and case matching on it. In Error case, i am updating port by adding 1 and retrying for port.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -731,12 +731,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "get-port"
-version = "4.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "888123007db34fbff15b5a347d46364dfbad531d6cb43de52cc0b62558f570e2"
-
-[[package]]
 name = "getrandom"
 version = "0.2.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1530,7 +1524,6 @@ dependencies = [
  "cfg-if",
  "console_error_panic_hook",
  "console_log",
- "get-port",
  "http",
  "leptos",
  "leptos_axum",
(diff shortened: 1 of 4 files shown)


srid (changes requested):
I should have clarified in the issue. I think it is best if you re-use a crate for this, rather than writing it by hand. Less code is better.

There is: https://lib.rs/crates/get-port

srid:
Or, we can do it without any new crates by trying port `0` if the default port is unavailable. cf. https://github.com/hyperium/tonic/issues/667#issuecomment-847847088

I wonder if port 0 works in Axum

1vipulgupta:
Even if we get Port we still need some computation part to get ip which is configured in `Cargo.toml`. or we want to run on `127.0.0.1` always ?

1vipulgupta:
Adding Port 0 is not working, it is giving `The web page at http://127.0.0.1:0/ might be temporarily down or it may have moved permanently to a new web address.`

srid (commented):
@1vipulgupta Just tried out locally, it doesn't work for me:

```
❯ nix run
warning: Git tree '/Users/srid/code/nix-browser' is dirty
thread 'main' panicked at 'error binding to 127.0.0.1:3000: error creating server listener: Address already in use (os error 48)', /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/hyper-0.14.27/src/server/server.rs:79:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

`just watch` to the left (using port 3000), and `nix run` to right run latter: 

image

I'd expect `nix run` to use a new port.

srid:
I also see this in `just watch`,

image

which probably means our approach is wrong here.

srid:
> I wonder if port 0 works in Axum

Actually I just confirmed that this does indeed work.

PR #31, opened 2023-08-14
Parse output of `nix --version`

### LINKED ISSUE #28: Parse Nix version
Currently the dashboard displays whatever `nix --version` returns, which at the moment is:

```
nix (Nix) 2.17.0
```

We just want to extract the "2.17.0". 

Basically implement the TODO here:

https://github.com/juspay/nix-browser/blob/35e75b3b492985f3e8cd2845404e5ba12fdc7e52/src/nix/info.rs#L24-L25

- [ ] Which Rust parser to use? Do we envision parsing more things in the future, and does that affect our choice now?
- [ ] Have the dashboard display the parsed Nix version

shivaraj-bh:
resolves #28

### DIFF
--- a/src/nix/version.rs
+++ b/src/nix/version.rs
@@ -1,0 +1,78 @@
+//! Rust module for `nix --version`
+use leptos::*;
+use serde::{Deserialize, Serialize};
+use std::fmt;
+#[cfg(feature = "ssr")]
+use tracing::instrument;
+
+/// Nix version as parsed from `nix --version`
+#[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub struct NixVersion {
+    major: u32,
+    minor: u32,
+    patch: u32,
+}
+
+/// Parse the string output of `nix --version` into a [NixVersion]
+pub fn parse_nix_version(version_str: &str) -> Option<NixVersion> {
+    let components: Vec<&str> = version_str.split_whitespace().collect();
+
+    if components.len() >= 2 {
+        let version = components[components.len() - 1];
+        let version_parts: Vec<&str> = version.split('.').collect();
+
+        if version_parts.len() == 3 {
+            let major = version_parts[0].parse().ok()?;
+            let minor = version_parts[1].parse().ok()?;
+            let patch = version_parts[2].parse().ok()?;
+
+            Some(NixVersion {
+                major,
+                minor,
+                patch,
+            })
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
+
+/// Get the output of `nix --version`
+#[cfg(feature = "ssr")]
+#[instrument(name = "version")]
+pub async fn run_nix_version() -> Result<NixVersion, ServerFnError> {
+    use tokio::process::Command;
+    let mut cmd = Command::new("nix");
+    cmd.arg("--version");
+    let stdout: Vec<u8> = crate::command::run_command(&mut cmd).await?;
+    let v = parse_nix_version(std::str::from_utf8(&stdout).unwrap()).unwrap();
+    Ok(v)
+}
+
+/// The HTML view for [NixVersion]
+impl IntoView for NixVersion {
+    fn into_view(self, cx: Scope) -> View {
+        view! { cx,
+            <div class="p-1 my-1 rounded bg-primary-50">
+                <pre>{self.major} . {self.minor} . {self.patch}</pre>
+            </div>
+        }
+        .into_view(cx)
+    }
+}
+
+/// The String view for [NixVersion]
+impl fmt::Display for NixVersion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}.{}.{}", self.major, self.minor, self.patch)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/nix/version.rs
```diff
@@ -0,0 +1,78 @@
...
+    patch: u32,
+}
+
+/// Parse the string output of `nix --version` into a [NixVersion]
+pub fn parse_nix_version(version_str: &str) -> Option<NixVersion> {
+    let components: Vec<&str> = version_str.split_whitespace().collect();
+
+    if components.len() >= 2 {
+        let version = components[components.len() - 1];
+        let version_parts: Vec<&str> = version.split('.').collect();
```
srid, line 22: Wouldn't it be simpler to use regex?

https://docs.rs/regex/latest/regex/

src/nix/version.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Get the output of `nix --version`
+#[cfg(feature = "ssr")]
+#[instrument(name = "version")]
+pub async fn run_nix_version() -> Result<NixVersion, ServerFnError> {
+    use tokio::process::Command;
+    let mut cmd = Command::new("nix");
+    cmd.arg("--version");
+    let stdout: Vec<u8> = crate::command::run_command(&mut cmd).await?;
+    let v = parse_nix_version(std::str::from_utf8(&stdout).unwrap()).unwrap();
```
srid, line 50: We should not use `unwrap`, [it panics](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap). Also, use `from_utf8_lossy` - since we don't care about utf8 errors in version strings.

src/nix/version.rs
```diff
@@ -0,0 +1,78 @@
...
+/// The String view for [NixVersion]
+impl fmt::Display for NixVersion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}.{}.{}", self.major, self.minor, self.patch)
+    }
+}
+
+#[cfg(feature = "ssr")]
+#[tokio::test]
+async fn test_run_nix_version() {
```
srid, line 75: Could you also add a test for `parse_nix_version`? Pass it a bunch of hardcoded version strings as input.

srid (changes requested):
Good start, just a few changes needed.

src/nix/version.rs
```diff
@@ -0,0 +1,78 @@
...
+    Ok(v)
+}
+
+/// The HTML view for [NixVersion]
+impl IntoView for NixVersion {
+    fn into_view(self, cx: Scope) -> View {
+        view! { cx,
+            <div class="p-1 my-1 rounded bg-primary-50">
+                <pre>{self.major} . {self.minor} . {self.patch}</pre>
+            </div>
```
srid, line 60: The IntoView for version string should just be a `<pre>`. 

The `<div>` element actually belongs to the NixConfig renderer. 

Also, why not `<pre>{self}</pre>`? Doesn't that use the `Display` instance?

shivaraj-bh: I have moved the `<div>` back to NixInfo renderer. `<pre>{self}</pre>` fails with stack overflow.

shivaraj-bh: This worked: `{format!("{}", self)}`

shivaraj-bh:
I have made the requested changes. 

I have a question:

Why am I getting this warning even though I am using `Regex` in `parse_nix_version` function?
```bash
warning: unused import: `regex::Regex`
 --> src/nix/version.rs:3:5
  |
3 | use regex::Regex;
  |     ^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

src/nix/version.rs
```diff
@@ -0,0 +1,94 @@
+//! Rust module for `nix --version`
+use leptos::*;
+use regex::Regex;
```
srid, line 3: Since this import is only used a function that is available server-side only:

```suggestion
#[cfg(feature = "ssr")]
use regex::Regex;
```

src/nix/version.rs
```diff
@@ -0,0 +1,94 @@
...
+#[tokio::test]
+async fn test_run_nix_version() {
+    let nix_version = run_nix_version().await.unwrap();
+    println!("Nix version: {}", nix_version);
+}
+
+#[cfg(feature = "ssr")]
+#[tokio::test]
+async fn test_parse_nix_version() {
+    let parsed_nix_version = parse_nix_version("nix (Nix) 2.13.0".to_string()).unwrap();
```
srid, line 78: Instead of `unwrap` (which will panic) just compare directly to Option, as in: `assert_eq!(parsed_nix_version, Some(NixVersion { ...})))`.

src/nix/version.rs
srid: > I have a question:
> 
> Why am I getting this warning even though I am using `Regex` in `parse_nix_version` function?

Because of this ^

shivaraj-bh:
Is it good to be merged now?

srid:
@shivaraj-bh FYI - https://github.com/juspay/nix-browser/commit/d48bc2b2a17db267925a651f7940293d5cb2c2fd

PR #35, opened 2023-08-15
Add --no-open; automatically open in web browser

1vipulgupta:


srid (changes requested):
As discussed in DM, we want something that resolves #13.

srid (approved):

PR #41, opened 2023-08-17
cli: add verbosity option (`-vv...`)

### LINKED ISSUE #36: CLI: Add verbosity flags
The CLI should take the following args:

- `--verbose` or `-v`: print DEBUG logs
- `-vv`: print TRACE logs
- `-q`: print WARN log and above (not INFO)

(Note: you can't use `-q` and `-v*` together; is it possible to enforce this constraint using `clap` library? Let's make `-q` a stretch goal, not necessary for first PR)

The tracing library has different [levels](https://docs.rs/tracing/latest/tracing/struct.Level.html). Currently, we set the log-level to INFO. But we want the user to control that.

It should also affect the frontend logging level.

1vipulgupta:
resolves #36

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -8,14 +8,13 @@ pub struct Args {
     /// Enabled by default if the app is running under `cargo leptos ...`
     #[arg(short = 'n', long = "no-open", default_value_t = in_cargo_leptos())]
     pub no_open: bool,
-    /// This flag alters the level log based on count of arg.
-    ///
-    /// 1 -> DEBUG, 2 -> TRACE, Default -> INFO
+    /// Be verbose in server logging (-v, -vv)
     #[arg(short = 'v', long = "verbose", action = clap::ArgAction::Count , default_value_t = 0)]
     pub verbose: u8,
 }
 
 impl Args {
+    /// Return the server log level
     pub fn log_level(&self) -> tracing::Level {
         match self.verbose {
             0 => tracing::Level::INFO,
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -12,6 +12,7 @@ pub struct Args {
     /// Enabled by default if the app is running under `cargo leptos ...`
     #[arg(short = 'n', long = "no-open", env = "NIX_BROWSER_NO_OPEN")]
     pub no_open: bool,
+
     /// The address to serve the application on
     ///
     /// Format: `IP_ADDRESS:PORT`
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -12,8 +12,18 @@ pub struct Args {
     #[arg(long = "vv", default_value_t = in_cargo_leptos())]
     pub vv: bool,
     ///This flag enables the DEBUG level log.
-    #[arg(short = 'v', long = "verbose", default_value_t = in_cargo_leptos())]
-    pub verbose: bool,
+    #[arg(short = 'v', long = "verbose", action = clap::ArgAction::Count , default_value_t = 0)]
+    pub verbose: u8,
+}
+
+impl Args {
+    pub fn log_level(&self) -> tracing::Level {
+        match self.verbose {
+            1 => tracing::Level::DEBUG,
+            2 => tracing::Level::TRACE,
+            _ => tracing::Level::INFO,
+        }
+    }
 }
 
 /// Whether the app is running under `cargo leptos ...`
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -60,9 +60,16 @@ async fn create_server() -> axum::Server<AddrIncoming, IntoMakeService<axum::Rou
...
+    let mut trace_level = tracing::Level::INFO;
+    if args.verbose {
+        trace_level = tracing::Level::DEBUG;
+    }
+    if args.vv {
+        trace_level = tracing::Level::TRACE;
+    }
     tracing_subscriber::fmt()
-        .with_max_level(tracing::Level::INFO) // TODO: --verbose should use DEBUG
+        .with_max_level(trace_level)
```
srid, line 72: You should also set the Axum logging level here,

https://github.com/juspay/nix-browser/blob/132ed807fa909b0eaa73871b28f0f6c255b50703/src/server.rs#L53-L54

src/cli.rs
```diff
@@ -8,6 +8,12 @@ pub struct Args {
     /// Enabled by default if the app is running under `cargo leptos ...`
     #[arg(short = 'n', long = "no-open", default_value_t = in_cargo_leptos())]
     pub no_open: bool,
+    ///This flag enables the TRACE and DEBUG level log.
+    #[arg(long = "vv", default_value_t = in_cargo_leptos())]
+    pub vv: bool,
+    ///This flag enables the DEBUG level log.
+    #[arg(short = 'v', long = "verbose", default_value_t = in_cargo_leptos())]
+    pub verbose: bool,
```
srid, line 16: Why _two_ flags? You don't have to implement this by hand, as `clap` already supports it. See here for an example,

https://github.com/DeterminateSystems/nix-installer/blob/329861feca73c00362bbe5004eb30fdb92ba398b/src/cli/arg/instrumentation.rs#L32-L34

src/server.rs
```diff
@@ -18,7 +18,7 @@ use crate::cli;
 
 /// Axum server main entry point
 pub async fn main(args: cli::Args) {
-    setup_logging();
+    setup_logging(&args);
```
srid, line 21: `setup_logging` should only take what it needs -- verbosity level -- not the whole args.

As a general rule, a function must take only the arguments it needs.

src/server.rs
```diff
@@ -60,9 +60,16 @@ async fn create_server() -> axum::Server<AddrIncoming, IntoMakeService<axum::Rou
     server
 }
 
-fn setup_logging() {
+fn setup_logging(args: &cli::Args) {
+    let mut trace_level = tracing::Level::INFO;
```
srid, line 64: You can write this without mutable variables. Tip: `if` is an expression, not a statement.

srid: But from an API standpoint you are better of `impl`'ing it on the Cli type itself. 

Like https://github.com/DeterminateSystems/nix-installer/blob/329861feca73c00362bbe5004eb30fdb92ba398b/src/cli/arg/instrumentation.rs#L46-L53

src/server.rs
srid: (Make sure to merge from `main` first, because I touched these lines today)

src/server.rs
1vipulgupta:
```
let trace_level = if verbose == 1 {
        tracing::Level::DEBUG
    } else if verbose == 2 {
        tracing::Level::TRACE
    } else {
        tracing::Level::INFO
    };
```

src/cli.rs
```diff
@@ -8,6 +8,21 @@ pub struct Args {
...
+    #[arg(short = 'v', long = "verbose", action = clap::ArgAction::Count , default_value_t = 0)]
+    pub verbose: u8,
+}
+
+impl Args {
+    pub fn log_level(&self) -> tracing::Level {
+        match self.verbose {
+            1 => tracing::Level::DEBUG,
+            2 => tracing::Level::TRACE,
+            _ => tracing::Level::INFO,
```
srid, line 23: What happens if the user passes, say, `-vvv`?

1vipulgupta: Updated case match with 0, 1, _

srid:
@1vipulgupta I just pushed a small refactoring of the logging code (to make them all part of the same module for legibility); could you merge from `main`?

srid (approved):

PR #44, opened 2023-08-19
Nix cache URL check should ignore trailing slash

vimalkumar:
substituters = https://cache.nixos.org https://nammayatri.cachix.org

The above config without trailing slash should be considered as correct.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1729,6 +1729,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "tracing-subscriber-wasm",
+ "url",
  "uuid",
  "wasm-bindgen",
 ]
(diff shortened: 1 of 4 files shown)


srid (commented):
Yea, we should do this (ignore the trailing slash). But this particular change will break on my machine which does use a slash:

```
❯ nix show-config | grep '^substituters'
substituters = https://cache.nixos.org/
```

We can perhaps improve on this PR by replacing the `String` in the below type,

https://github.com/juspay/nix-browser/blob/a83e8b69875b974b25adeaeda49ec0b0ab4d6eef/src/nix/config.rs#L16

... with the `Url` type from https://docs.rs/url/latest/url/ and then doing equality on them.

vimalkumar:
Since it does a val.contains() and not an exact string match, I thought it would work for both - URLs with and without a trailing slash. 

I think you have already taken care of it. Thanks.

PR #54, opened 2023-09-04
e2e playwright test

### LINKED ISSUE #11: Add end to end testing

divinenaman:
fixes #11 using playwright

- [x] Initial impl
- [x] Github Actions CI check
- [x] VSCode IDE support w/ instructions


src/server.rs
srid, line 1: We shouldn't be modifying the API endpoints just for e2e test. The changes to `server.rs` should be reverted.

e2e-playwright/tests/nix-version.test.js
```diff
@@ -0,0 +1,16 @@
+// @ts-check
+const { test, expect } = require('@playwright/test');
+
+test('check nix version', async ({ page, request }) => {
+  await page.goto('/info');
+  const nixVersion = await page.locator(":text('Nix Version') + div").textContent();
+
+  const apiResponse = await request.get(`/api/data/nix-info`);
```
srid, line 8: We want to test that **some** version is shown in the browser, and not that the _correct_ version is displayed. Thus, there is no need to poll any API.

Cargo.lock
srid, line 1: Why does this PR have changes to `Cargo.lock`?

srid (commented):
@divinenaman The PR is coming along well. I've left some comments. Could you add Github Actions workflow to run the e2e tests in CI?

e2e-playwright/flake-module.nix
```diff
@@ -0,0 +1,41 @@
+{ self, lib, inputs, ... }:
+{
+  perSystem = { config, self', pkgs, system, ... }:
+    {
+      # e2e test service using playwright
+      process-compose.cargo-e2e-playwright-test =
```
srid, line 6:
```suggestion
      process-compose.e2e-playwright-test =
```

Cargo.lock
divinenaman: it was giving errors

divinenaman:
@srid is there a way to mock, github actions ? I remember it took around 10 commits last time I did something similar

srid:
> @srid is there a way to mock, github actions ? I remember it took around 10 commits last time I did something similar

I don't know, but you do have write access to this repo, so you can push this branch to this repo (rather than your fork).

justfile
```diff
@@ -17,7 +17,11 @@ alias w := watch
 
 # Run tests (backend & frontend)
 test:
-    cargo watch -- cargo leptos test
+    cargo-test
```
srid, line 20: Why was `cargo watch -- cargo leptos test` replaced?

e2e-playwright/tests/nix-version.test.js
srid, line 1: @divinenaman How is the IDE experience in VSCode like when editing this file?

justfile
srid: Oh, I just realized that tests were broken on `main`. I just fixed it in 855e964924bf94ea78aad04b16add08512db7e34 - you want to merge that.

e2e-playwright/tests/nix-version.test.js
divinenaman: @srid  I am just happy with onSaveFormat, for me it's good enough

srid: If we introduced `package.json` and the like, would it provide an improved experience (like autocompletion, and hover tooltips)?

srid: Let's do that, and explain that in `e2e-playwright/README.md`. We want good IDE experience for writing tests as well.

Cargo.lock
srid: @divinenaman The changes to `Cargo.lock` should be reverted. If you are seeing any errors, try to resolve them by finding the root cause - or post it here for help.

src/server.rs
```diff
@@ -5,7 +5,11 @@ use crate::app::App;
 use axum::response::Response as AxumResponse;
 use axum::routing::IntoMakeService;
 use axum::{body::Body, http::Request, response::IntoResponse};
-use axum::{routing::post, Router};
+use axum::{
+    http::StatusCode,
+    routing::{get, post},
+    Router,
+};
```
srid, line 12: Should be reverted

Cargo.lock
srid: Note: you can do this using `git show main:Cargo.lock > Cargo.lock`

e2e-playwright/tests/nix-version.test.js
srid: @divinenaman How do I get IDE support in vscode? Can you explain the steps?

This is what I see:

image

Cargo.lock
divinenaman: @srid reverted the changes, working as expected

e2e-playwright/tests/nix-version.test.js
srid: I was able to fix this by running `npm i` as well as adding the `package-lock.json`: https://github.com/juspay/nix-browser/commit/021e936e2468e88b0a3165696de46d1d4a322e38

However, it broke `nix run`:

image

@divinenaman Can you figure out a solution here? This is the only thing to do before merging this PR.

divinenaman: @srid I am not able to re-create, I did `nix-store --gc` to clear my nix store but tests passed for me, any hints to recreate ?

divinenaman: @srid can you once try now, added playwright browser path

srid: > also remove the local node-modules

It works without local node-modules.  But this local node-modules (created `npm i`) exists solely so that VSCode IDE support will work. So how would you make both work at the same time?

srid: > You have two different versions of @playwright/test. This usually happens
> when one of the dependencies in your package.json depends on @playwright/test.

This seems to be the culprit.

srid: I think I have a solution, hold on.

srid (approved):
Awesome, everything's good!

PR #57, opened 2023-09-06
Add CLI for nix-health

srid:
The idea is to eventually supplant https://github.com/srid/nix-health

- [x] Full console reporting
- [x] `nix run` invocation
- [ ] publish crate


srid:
image

Issue #58, opened 2023-09-08
`nix_health`: check if rosetta is in use


srid:
This will be a new kind of check, as it must happen only on ARM macOS.

srid:
With #64, we can make this check run only on ARM macOS, ignoring other platforms, because the `check` function now returns an `Option`.

PR #62, opened 2023-09-12
`nix_health`: `trusted_users` check

### LINKED ISSUE #37: `nix_health`: check $USER is in trusted-users
To catch scenarios like this, we need a health check that checks the user is in trusted-user. 

```
❯ nix run nixpkgs#cachix use nammayatri

This user doesn't have permissions to configure binary caches.

You can either:

a) Run the same command as root to configure them globally.

b) Run the following command to add your user as trusted
   and then try again:

  echo "trusted-users = root srid" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon
```

To implement this, you can add a new field to the `NixConfig` struct called `trusted_users`, and add a new health check (like #32) to test the current user is in it.

shivaraj-bh:
Resolves #37 

TODO:
- [x] Change suggestion based on `os` from `SysInfo`


srid (commented):
Is the `system_rs` crate going to be useful in general sense? Given that it contains only information needed for `nix_health`, I'd think not. In that case, `SysInfo` should be part of `nix_rs` itself (presuming it provides information about the environment in which Nix is installed).

(Bear in mind that any local crate that `nix_health` depends will have to be published to crates.io, since we publish `nix_health` itself: https://crates.io/crates/nix_health)

crates/system_rs/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "0.1.0"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cfg-if.workspace = true
+os_info = "3.7.0"
```
srid, line 13: Can this crate be used for #58 as well?

shivaraj-bh: I will verify if that's possible.

shivaraj-bh:
> Is the system_rs crate going to be useful in general sense?
I thought it could add on top of whatever `os_info` already provides and hence I was considering a separate crate. Maybe I should just stick to moving it to `nix_rs` for now and when we have enough information about the system and if the community needs such a crate we can separate it out.

shivaraj-bh:
Whether I am in rosetta or not, I get the following output from the `os_info` crate:

```bash
OS information: Mac OS 13.5.0 [64-bit]
Type: Mac OS
Version: 13.5.0
Edition: None
Codename: None
Bitness: 64-bit
Architecture: Some("arm64")
```

shivaraj-bh:
@srid I have added the condition to check whether the mac is configured using `nix-darwin`. I don't have a mac that uses it atm, it would be great if you could confirm that it works fine.

srid:
> @srid I have added the condition to check whether the mac is configured using `nix-darwin`. I don't have a mac that uses it atm, it would be great if you could confirm that it works fine.

Alright, I'll take over from here.

srid (approved):

shivaraj-bh:
@srid Went through the refactored code, was very helpful. Thanks!

PR #69, opened 2023-09-15
gh-actions: Run after garnix finishes

srid:
Along with #68 this seeks to make e2e CI in github faster.

- [ ] make it work


srid:
Doesn't work, asked in https://github.com/orgs/community/discussions/67216

PR #79, opened 2023-10-03
Leptos -> Dioxus Desktop

srid:
- [x] Port leptos code
- [ ] Run `dx fmt` as part of treefmt https://github.com/juspay/nix-browser/issues/81
- [x] nixify packages
- [x] Avoid unnecessary clones
- [x] Nav bar: active/inactive status
- ~~[ ] component should take signals~~ (too early to do this)
- [x] reintroducing loading status everywhere
- [x] final refactor


srid:
It is unclear how application state should be handled. Will explore first in https://github.com/srid/dioxus-desktop-template

PR #87, opened 2023-10-18
Open local flake with a `FolderDialogButton`

### LINKED ISSUE #85: Add "open directory" button to open a local flake
Add a "Open directory" button here:

image

This should ideally bring up the user's [native file-open dialog](https://en.wikipedia.org/wiki/File_dialog) allowing them to select their local project directory. Then the input box is replaced with that local path (eg: `/Users/user/Projects/nammayatri`) which then triggers a reload of the flake info.

## Implementation notes

- Perhaps we can uses Tauri's API to do this: https://tauri.app/v1/api/js/dialog/#open
  - ~~The Rust API seems to miss the 'open' dialog which we may have to contribute? https://docs.rs/tauri/1.5.2/tauri/api/dialog/index.html~~
  - Rust API: https://docs.rs/tauri/latest/tauri/api/dialog/struct.FileDialogBuilder.html

shivaraj-bh:
Resolves #85

### DIFF
--- a/src/app/flake.rs
+++ b/src/app/flake.rs
@@ -38,8 +38,28 @@ pub fn Flake(cx: Scope) -> Element {
                     fut.restart();
                 }
             }
-            // TODO: Use the FileOpenDialog handler
-            div { class: "ml-2", FileExplorerButton { handler: move |_| { fut.restart() } } }
+            div { class: "ml-2 flex flex-col",
+                FileExplorerButton {
+                    handler: move |evt: Event<FormData>| {
+                        match &evt.files {
+                            Some(file_engine) => {
+                                if let Some(flake_path) = file_engine.files().first() {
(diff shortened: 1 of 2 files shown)


srid:
Just found a small bug: when you click on the 'cancel' button, the handler is still invoked (tracing an error in the process).

shivaraj-bh:
Based on how FormData's [struct](https://github.com/DioxusLabs/dioxus/blob/c7963a03440d5a050bf229f91665d60a0d108a8a/packages/html/src/events/form.rs#L10-L24) is defined, I don't see a straightforward way to catch this; our best bet could be to change the [error](https://github.com/juspay/nix-browser/blob/5bd385f5c11d399f42b02c395d1d93ad810f6e8c/src/app/widget.rs#L64) to `warning` and say `"user cancelled/unable to get the selected path"`

PR #98, opened 2023-10-31
Persist `Signal` state to disk (recently visited flakes now persist)

### LINKED ISSUE #96: Introduce persistent application state
Use a ~~sqlite db~~ dioxus persistance to store application state which includes:

- [x] Recent flakes accessed https://github.com/juspay/nix-browser/pull/98
- [x] Flake output structure https://github.com/juspay/nix-browser/pull/101
   - Once we "cache" the outputs in the DB, it can be shown without having re-evaluate the flake again. This can be useful for large projects with IFD, like nammayatri, or even nixpkgs. The UI will open ready to do search for packages, etc.
- [ ] UI indicator for stale data. If a flake is out of date, encourage user to update it. Then update the SystemTime in the cache.
- User settings if any (in the future)

srid:
For #96 by leveraging https://github.com/DioxusLabs/dioxus-std/pull/17

- [x] "Recent flakes" should work
- [x] Add a default list of flakes (if recent flakes is empty)


src/app/state.rs
```diff
@@ -65,6 +72,16 @@ impl Action {
...
 impl AppState {
+    fn new(cx: Scope) -> Self {
+        tracing::debug!("🔨 Creating AppState default value");
+        let recent_flakes =
+            storage::<LocalStorage, _>(cx, "recent_flakes".to_string(), FlakeUrl::suggestions);
+        AppState {
+            recent_flakes,
+            ..AppState::default()
+        }
+    }
```
srid, line 83: This (`storage::<LocalStorage, _>`) is how we create a dioxus `Signal` that is backed by a persistent storage behind the scenes, thus making the signal itself persistent across app restarts!

PR #101, opened 2023-11-08
Cache retrieved flakes on disk

srid:
Flakes are fetched once and cached on disk, so as to persist across app restarts. To refetch, the user can hit the 'refresh' button.

- [x] dioxus-std storage deserialization issues https://github.com/DioxusLabs/dioxus-std/pull/17#pullrequestreview-1719137114
- [ ] trim cache size?
- [ ] refactor build_network logic


srid:
>  refactor build_network logic

Will do this in subsequent PR.

PR #110, opened 2023-12-13
Make `nix-health` trusted-users check recognize groups and all-users

### LINKED ISSUE #104: Trusted Users can be a group of which you are part
Putting yourself into a group of trusted users is also valid.
```
❌ Trusted Users
   trusted-users = @wheel
   User 'xyz' not present in trusted_users
   Add `nix.trustedUsers = [ "root" "xyz" ];` to your nixos configuration
```

This is in the nix manual:
image

aravindgopall:
Resolves #104

### DIFF
--- a/crates/nix_rs/src/env.rs
+++ b/crates/nix_rs/src/env.rs
@@ -39,7 +39,9 @@ impl NixEnv {
             );
             let total_disk_space = to_bytesize(get_nix_disk(&sys)?.total_space());
             let total_memory = to_bytesize(sys.total_memory());
-            let output = Command::new("groups").output()?;
+            let output = Command::new("groups")
+                .output()
+                .map_err(NixEnvError::GroupsError)?;
             let group_info = &String::from_utf8_lossy(&output.stdout);
             Ok(NixEnv {
                 current_user,
@@ -197,7 +199,7 @@ pub enum NixEnvError {
     EnvVarError(#[from] std::env::VarError),
 
     #[error("Failed to fetch groups: {0}")]
-    IOError(#[from] std::io::Error),
+    GroupsError(std::io::Error),
 
     #[error("Unable to find root disk or /nix volume")]
     NoDisk,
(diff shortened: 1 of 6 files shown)


crates/nix_rs/src/info.rs
```diff
@@ -10,6 +11,7 @@ pub struct NixInfo {
     pub nix_version: NixVersion,
     pub nix_config: NixConfig,
     pub nix_env: NixEnv,
+    pub group_info: String,
```
srid, line 14: The user's group should reside in the same struct that contains the user's username, viz. `NixEnv`:

https://github.com/juspay/nix-browser/blob/83746cc2e0e720b1f38f6951ae0fa89c98c1e4fb/crates/nix_rs/src/env.rs#L11-L13

srid: Also, the type should be a list not a string, viz.:

```rust
pub struct NixEnv {
  pub user_groups: Vec<String>
```

srid (changes requested):

crates/nix_rs/src/env.rs
```diff
@@ -29,15 +32,22 @@ impl NixEnv {
         use sysinfo::{DiskExt, SystemExt};
         tracing::info!("Detecting Nix environment");
         let os = OS::detect().await;
-        tokio::task::spawn_blocking(|| {
+        let output = Command::new("groups").output().await.unwrap();
```
srid, line 35: `unwrap` should generally not be used (it will crash the app). This function already returns a `Result`, so you want to marshall this error in a `NixEnvError`.

See how we do this for `current_user`.

srid (changes requested):

srid (approved):
LGTM. 

The `groups` variable however can be modified to strip out that `@` such that the subsequent membership check is simplified.

I'll take a look at this once I'm back from Montreal.

srid (approved):
@aravindgopall I've refactored and addressed a few things. Check out the commit log for details.

srid:
I like how now the whole logic is compartmentalized into the `TrustedUserValue` enum, which serde deserializes into directly from `nix show-config --json`.

PR #111, opened 2023-12-19
fixed intel build

1vipulgupta:


flake.nix
```diff
@@ -71,13 +71,16 @@
...
             just
             nixci
             # For when we start using Tauri
             cargo-tauri
             trunk
-          ];
+          ]) ++ lib.optionals pkgs.stdenv.isDarwin (
+            [ pkgs.darwin.apple_sdk_11_0.frameworks.IOKit
+              pkgs.darwin.apple_sdk_11_0.frameworks.Security
+            ]);
```
srid, line 83: This should go in rust.nix (`rustBuildInputs`). Also, CI fails due to not having run autoformatter.

srid (changes requested):

flake.nix
1vipulgupta: Adding in `rustBuildInputs` is causing nix run failing. Updated and added to `devShells.rust`

srid: > causing nix run failing

Please post the full error.

---

devShells.rust already uses rustBuildInputs transitively, so the devShell will now have duplicate IOKit and Security frameworks.

1vipulgupta: `error occurred: Command env -u IPHONEOS_DEPLOYMENT_TARGET "clang" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "--target=x86_64-apple-darwin" "-Wall" "-Wextra" "-o" "/private/tmp/nix-build-nix-browser-deps-0.1.0.drv-0/source/target/release/build/tao-fde661b9405e71cb/out/src/platform_impl/macos/carbon_hotkey/carbon_hotkey_binding.o" "-c" "src/platform_impl/macos/carbon_hotkey/carbon_hotkey_binding.c" with args "clang" did not execute successfully (status code exit status: 1).`

srid:
`main` branch builds on Intel mac (https://github.com/juspay/nix-browser/blob/main/.github/workflows/ci.yaml) and I've confirmed that the app runs locally in Rosetta.

Issue #120, opened 2024-03-13
`nix-health`: Check that direnv doesn't come from homebrew


srid:
For eg., `/usr/local/bin/direnv`. homebrew installs fucked up direnv that is glitchy on VSCode.

PR #123, opened 2024-03-19
fix(nix-health): use `direnv status --json` & create `direnv` crate

### LINKED ISSUE #122: `nix-health`: `direnv allow` check failure
Screenshot 2024-03-19 at 12 15 41 AM

The check above fails even though `direnv allow` has been run:
Screenshot 2024-03-19 at 12 15 54 AM

Here 0 means `Allowed`, for explanation, see: https://github.com/direnv/direnv/pull/1142#issuecomment-1848677532

shivaraj-bh:
resolves #122 

* fix the broken activation (`direnv allow`) check

### DIFF
--- a/crates/nix_health/src/check/direnv.rs
+++ b/crates/nix_health/src/check/direnv.rs
@@ -22,7 +22,6 @@ impl Default for Direnv {
 }
 
 impl Checkable for Direnv {
-    // TODO: This code flow is confusing; refactor for legibility.
     fn check(&self, _nix_info: &info::NixInfo, flake_url: Option<FlakeUrl>) -> Vec<Check> {
         let mut checks = vec![];
         if !self.enable {
@@ -30,25 +29,20 @@ impl Checkable for Direnv {
         }
 
         let direnv_install_result = direnv::DirenvInstall::detect();
+        checks.push(install_check(&direnv_install_result, self.required));
 
-        let direnv_install_check = install_check(&direnv_install_result, self.required);
-        let direnv_installed = direnv_install_check.result.green();
-        checks.push(direnv_install_check);
-
-        if !direnv_installed {
-            return checks;
-        }
-
-        // FIXME: Avoid unwrap, by refactoring code flow.
-        let direnv_install = direnv_install_result.as_ref().unwrap();
-
-        // If direnv is installed, check for version and then allowed_check
-        // This check is currently only relevant if the flake is local and an `.envrc` exists.
-        match flake_url.as_ref().and_then(|url| url.as_local_path()) {
-            None => {}
-            Some(local_path) => {
-                if local_path.join(".envrc").exists() {
-                    checks.push(allowed_check(direnv_install, local_path, self.required));
+        match direnv_install_result.as_ref() {
+            Err(_) => return checks,
+            Ok(direnv_install) => {
+                // If direnv is installed, check for version and then allowed_check
+                // This check is currently only relevant if the flake is local and an `.envrc` exists.
+                match flake_url.as_ref().and_then(|url| url.as_local_path()) {
+                    None => {}
+                    Some(local_path) => {
+                        if local_path.join(".envrc").exists() {
+                            checks.push(allowed_check(direnv_install, local_path, self.required));
+                        }
+                    }
                 }
             }
         }
--- a/crates/nix_health/src/check/direnv.rs
+++ b/crates/nix_health/src/check/direnv.rs
@@ -137,6 +137,7 @@ fn allowed_check(
 mod direnv_crate {
     use semver::Version;
     use serde::{Deserialize, Serialize};
+    use serde_repr::{Deserialize_repr, Serialize_repr};
     use std::path::PathBuf;
 
     /// Information about a local direnv installation
@@ -222,7 +223,9 @@ mod direnv_crate {
     impl DirenvState {
         /// Check if the .envrc file is allowed
         fn is_allowed(&self) -> bool {
-            self.found_rc.as_ref().map_or(false, |rc| rc.allowed == 0)
+            self.found_rc
+                .as_ref()
+                .map_or(false, |rc| rc.allowed == AllowedStatus::Allowed)
         }
     }
 
@@ -230,12 +233,20 @@ mod direnv_crate {
     /// TODO: Represent 0/1/2 values in a 3-value enum
     #[derive(Debug, Serialize, Deserialize, Clone)]
     struct DirenvRC {
-        /// Can be 0, 1 or 2
-        /// 0: Allowed
-        /// 1: NotAllowed
-        /// 2: Denied
-        allowed: u32,
+        allowed: AllowedStatus,
         /// Path to the .envrc file
         path: PathBuf,
+    }
+
+    #[derive(Serialize_repr, Deserialize_repr, PartialEq, Debug, Clone)]
+    #[repr(u8)]
+    /// Can be 0, 1 or 2
+    /// 0: Allowed
+    /// 1: NotAllowed
+    /// 2: Denied
+    enum AllowedStatus {
+        Allowed = 0,
+        NotAllowed = 1,
+        Denied = 2,
     }
 }
--- a/crates/nix_health/src/check/direnv.rs
+++ b/crates/nix_health/src/check/direnv.rs
@@ -31,9 +31,9 @@ impl Checkable for Direnv {
             return checks;
         }
 
-        let direnv_install = DirenvInstall::detect();
+        let direnv_install_result = DirenvInstall::detect();
 
-        let direnv_install_check = install_check(&direnv_install, self.required);
+        let direnv_install_check = install_check(&direnv_install_result, self.required);
         let direnv_installed = direnv_install_check.result.green();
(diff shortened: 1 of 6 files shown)


crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +93,74 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
 
-impl DirenvInstall {
-    /// Detect user's direnv installation
+#[derive(Debug, Serialize, Deserialize, Clone)]
+struct DirenvConfig {
+    #[serde(rename = "ConfigDir")]
+    config_dir: PathBuf,
+    /// Path to the direnv binary
+    #[serde(rename = "SelfPath")]
+    self_path: PathBuf,
```
shivaraj-bh, line 126: This will be useful to also solve #120

crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +93,74 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
+fn is_direnv_allowed_on(project_dir: &std::path::Path) -> anyhow::Result<bool> {
+    let output = std::process::Command::new("direnv")
+        .args(["status", "--json"])
+        .current_dir(project_dir)
+        .output()?;
+    if output.status.success() {
+        let out = String::from_utf8_lossy(&output.stdout);
+        let status = DirenvStatus::from_json(&out)?;
+        // rc.allowed == 0 means that the .envrc is allowed, see: https://github.com/direnv/direnv/pull/1142#issuecomment-1848677532
+        Ok(status.state.found_rc.map_or(false, |rc| rc.allowed == 0))
```
srid, line 107: You can move this to a method on the struct, viz.:

```rust
impl DirenvStatus {
  fn is_allowed(self) -> bool { ... }
}
```

This way the caller doesn't have to care about the implementation; only the semantics.

srid: Actually it should be on `DirenvState`

crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +93,74 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
-    pub bin_path: PathBuf,
+/// [Check] if direnv was already allowed in [project_dir]
 
-    /// Contents of `direnvrc`
-    pub direnv_config: Option<String>,
+fn is_direnv_allowed_on(project_dir: &std::path::Path) -> anyhow::Result<bool> {
+    let output = std::process::Command::new("direnv")
+        .args(["status", "--json"])
+        .current_dir(project_dir)
+        .output()?;
```
shivaraj-bh, line 102: `direnv status —json` is only possible from direnv >= 2.33.0, so older versions will not be supported.

I am wondering if it would be better to introduce a `version_check`

crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +93,74 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
+    config: DirenvConfig,
+    state: DirenvState,
 }
 
-impl DirenvInstall {
-    /// Detect user's direnv installation
+#[derive(Debug, Serialize, Deserialize, Clone)]
+struct DirenvConfig {
+    #[serde(rename = "ConfigDir")]
+    config_dir: PathBuf,
```
srid, line 123: Could you add a `///` description here? What is this directory?

crates/nix_health/src/check/direnv.rs
srid: If it doesn't affect our internal projects like Euler, we could just drop support for old direnv to keep the code simple.

srid: But yes in that case, a version check is probably still useful, so we can tell the user (as red/orange message) in that case that their direnv version is too old.

shivaraj-bh: The file might get rather long if I have to create another struct for version. It won’t if I use the semver crate and let it parse the version.

srid: Better to reuse external crate when possible. 

https://docs.rs/semver/1.0.22/semver/

I wonder why we didn't use it here:

https://github.com/juspay/nix-browser/blob/358f3c444918bc4a577e7e06341173c40e0038e1/crates/nix_rs/src/version.rs#L11-L15

shivaraj-bh:
Here’s a handy screen grab for what it looks like currently:
Screenshot 2024-03-20 at 12 09 36 AM

crates/nix_health/src/check/direnv.rs
```diff
@@ -71,14 +80,45 @@ fn install_check(required: bool) -> Check {
...
+                suggestion,
+            },
+            Ok(None) => CheckResult::Red {
+                msg: "Unable to parse direnv version".to_string(),
+                suggestion,
+            },
+            Err(e) => CheckResult::Red {
+                msg: format!("Unable to locate direnv: {}", e),
+                suggestion,
+            },
```
srid, line 108: Shouldln't "Direnv Installation" check happen *before* the version check? Then this clause becomes obviated.

crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +133,99 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
+    #[serde(rename = "foundRC")]
+    found_rc: Option<DirenvRC>,
+    /// Information about the .envrc that is currently allowed using `direnv allow`
+    #[serde(rename = "loadedRC")]
+    loaded_rc: Option<DirenvRC>,
+}
+
+impl DirenvState {
+    /// Check if the .envrc file is allowed
+    fn is_found_rc_allowed(&self) -> bool {
```
srid, line 194:
```suggestion
    fn is_allowed(&self) -> bool {
```

Why? Because function name should reflect semantics not implementation.

crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +133,99 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
+        Ok(status.state.is_found_rc_allowed())
+    } else {
+        anyhow::bail!("Unable to run direnv status --json: {:?}", output.stderr)
     }
 }
 
-/// Check if direnv was already activated in [project_dir]
-
-pub fn is_direnv_active_on(project_dir: &std::path::Path) -> anyhow::Result<bool> {
+fn direnv_version() -> anyhow::Result<String> {
```
srid, line 152: I think this should return `Semver` instead of `String`. 

Function interface should be as smaller as possible.

shivaraj-bh:
Much better!
Screenshot 2024-03-20 at 12 29 57 AM

crates/nix_health/src/check/direnv.rs
```diff
@@ -93,71 +133,99 @@ fn activation_check(local_flake: &std::path::Path, required: bool) -> Check {
...
-impl DirenvInstall {
-    /// Detect user's direnv installation
+/// [Check] if direnv was already allowed in [project_dir]
 
-    pub fn detect() -> anyhow::Result<Self> {
-        let bin_path = which::which("direnv")?;
-        let output = std::process::Command::new(&bin_path)
-            .arg("status")
-            .output()?;
+fn is_direnv_allowed_on(project_dir: &std::path::Path) -> anyhow::Result<bool> {
```
srid, line 140: This can be moved as a method of `DirenvInstall` struct, no? This way we can run direnv using the exact detected path (`bin_path`) instead of searching $PATH again.

crates/nix_health/src/check/direnv.rs
srid: Likewise, this should probably be a method of `DirenvInstall`.

crates/nix_health/src/check/direnv.rs
```diff
@@ -31,17 +33,33 @@ impl Checkable for Direnv {
...
+        let direnv_install_check = install_check(&direnv_install, self.required);
         let direnv_installed = direnv_install_check.result.green();
         checks.push(direnv_install_check);
 
-        if direnv_installed {
+        if !direnv_installed {
+            return checks;
+        }
+
+        let direnv_version = version_check(direnv_install.as_ref().unwrap());
```
shivaraj-bh, line 46: After moving `direnv_version` and `is_direnv_allowed_on` to impl of `DirenvInstall`, I am passing the `direnv_install` object around, that doesn’t look nice.

srid: > I am passing the direnv_install object around, that doesn’t look nice.

That is not a problem in and of itself (structs are meant to be passed around anyway), unless you are referring to something else - such as the resulting code being complex.

I'll take a look, in a bit.

srid: > that doesn’t look nice.

I think you are mostly referring to the parts where I added TODO/FIXME in the code. But I want to improve the various `*_check` functions as well.

srid (approved):

srid:
https://crates.io/crates/direnv

PR #124, opened 2024-03-21
nix_health: check that direnv isn't installed globally

### LINKED ISSUE #120: `nix-health`: Check that direnv doesn't come from homebrew

shivaraj-bh:
resolves #120

### DIFF
--- a/crates/nix_health/src/check/direnv.rs
+++ b/crates/nix_health/src/check/direnv.rs
@@ -63,7 +63,7 @@ fn install_check(
             direnv_install_result.as_ref().ok().map(|s| &s.bin_path)
         ),
         result: match direnv_install_result {
-            Ok(direnv_install) => is_global(direnv_install),
+            Ok(direnv_install) => is_path_under_nix_store(direnv_install),
             Err(e) => CheckResult::Red {
                 msg: format!("Unable to locate direnv ({})", e),
                 suggestion: "Install direnv <https://nixos.asia/en/direnv#setup>".to_string(),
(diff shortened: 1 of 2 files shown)


crates/nix_health/src/check/direnv.rs
```diff
@@ -73,6 +73,20 @@ fn install_check(
     }
 }
 
+/// Verify that direnv was not installed globally, under `/usr`.
+fn is_global(direnv_install: &direnv::DirenvInstall) -> CheckResult {
+    let usr_path = std::path::Path::new("/usr");
+    if direnv_install.bin_path.starts_with(usr_path) {
```
srid, line 79: A better approach would be to resolve the binary symlink and check if it is in nix store (`/nix/store`):

```
❯ readlink $(which direnv)
/nix/store/birqvvy5rw3jkpy7n59nk1yk8wr3l217-home-manager-path/bin/direnv
```

This will then automatically handle the case of direnv being installed else, like `/opt`, on some systems.

srid: `fn is_path_under_nix_store(path: ...) -> bool`

shivaraj-bh: > A better approach would be to resolve the binary symlink 

added `canonical_path` field in `DirenvInstall` that stores that path without any symbolic links

crates/nix_health/src/check/direnv.rs
```diff
@@ -74,16 +74,16 @@ fn install_check(
 }
 
 /// Verify that direnv was not installed globally, under `/usr`.
-fn is_global(direnv_install: &direnv::DirenvInstall) -> CheckResult {
-    let usr_path = std::path::Path::new("/usr");
-    if direnv_install.bin_path.starts_with(usr_path) {
+fn is_path_under_nix_store(direnv_install: &direnv::DirenvInstall) -> CheckResult {
```
srid, line 77: Function interface should be as small in scope as possible: it should only take values it actually uses. Here, `direnv_install` is not used in its entirety, only `direnv_install.canonical_path` is. Indeed, the function name contains the word "path"; thus it makes more sense to pass this very path rather than some other object.

crates/nix_health/src/check/direnv.rs
```diff
@@ -74,16 +74,16 @@ fn install_check(
 }
 
 /// Verify that direnv was not installed globally, under `/usr`.
```
srid, line 76: Comment doesn't match function semantics.

srid (commented):
Alright, I'll test, change a few things if necessary and merge.

PR #158, opened 2024-07-17
feat: Statically link libiconv on `aarch64-darwin`; Test static binary in CI

### LINKED ISSUE #149: Static binary for `omnix-cli`
- Have `packages.omnix-cli` be statically built, both for Linux and macOS (should work on recent Mac systems we use at @juspay)
- Build it in CI, and upload it somewhere (GitHub release?) for people to download.

With `om` being a static binary it can even introspect (https://github.com/juspay/nix-health/issues/6) and do Nix install for the user.

Static binary status:
- [x] Linux: https://github.com/juspay/omnix/pull/185
- [ ] macOS: https://github.com/juspay/omnix/pull/158

shivaraj-bh:
Note: macOS system libraries will not be statically linked

resolves #149 

## Before
```sh
❯ otool -L ./result/bin/om
./result/bin/om:
        /System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 1109.60.2)
        /System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 59754.60.13)
        /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1770.255.0)
        /nix/store/0rsh55ighfv3fxidf3xz8vl3abiika0i-libiconv-99/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
```

## After
```sh
❯ otool -L ./result/bin/om
./result/bin/om:
        /System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 1109.60.2)
        /System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 59754.60.13)
        /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1770.255.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
```

In `After`, `libiconv` is statically linked.

TODO:
- [x] Use libiconv static on darwin, after https://github.com/NixOS/nixpkgs/pull/322497 reaches `nixpkgs-unstable` (Also see: https://github.com/NixOS/nixpkgs/issues/320448#issuecomment-2213814702)
- [x] https://github.com/juspay/omnix/pull/158#issuecomment-2233484730
~~- [ ] Consider how #166 will impact this.~~ (see https://github.com/juspay/omnix/pull/158#issuecomment-2244086640)


shivaraj-bh:
macOS still has a `dylib` dependency on `libiconv` apart from system libraries (see TODO in the description for solution):

```sh
❯ otool -L ./result/bin/om
./result/bin/om:
        /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1770.255.0)
        /nix/store/0rsh55ighfv3fxidf3xz8vl3abiika0i-libiconv-99/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
```

srid:
We should also test in Github Actions CI that the static binary runs (without Nix installed) on both platforms.

srid:
> We should also test in Github Actions CI (public runner, not self-hosted) that the static binary runs (without Nix installed) on both platforms.

Down the line, we may even re-use #160 tests to use the static binary rather than the cargo bin.

srid:
> * Consider how [Use flake-schemas #166](https://github.com/juspay/omnix/pull/166) will impact this.

Just added this to the todo list.

shivaraj-bh:
> > * Consider how [Use flake-schemas #166](https://github.com/juspay/omnix/pull/166) will impact this.
> 
> 
> 
> Just added this to the todo list.

I guess it should be fine if we wrap static nix instead of the default one.

srid:
There's also [`nix bundle`](https://nix.dev/manual/nix/2.22/command-ref/new-cli/nix3-bundle) but I don't know if it (or a variation of it) works on macOS.

srid:
I think we should gets this PR into `main` anyway, without considering the `nix-flake-schemas` dependency:

> [ ] Consider how [Use flake-schemas #166](https://github.com/juspay/omnix/pull/166) will impact this.

This PR will reduce the closure size of `nix run ... ` when retrieved from cachix (#172), and `omnix` will continue to work as long as `nix-flake-schemas` is in the closure of `omnix`.

srid:
@shivaraj-bh Let's get the Linux static binary part merged to `main` before we do the same for macOS. This will help with switching GitHub actions from `nixci` to `om ci`.

shivaraj-bh:
image

shivaraj-bh:
image

.github/workflows/ci.yaml
```diff
@@ -37,6 +37,15 @@ jobs:
...
         run: nix run .#check-closure-size
+      - name: nix build
+        if: matrix.system != 'x86_64-darwin'
+        run: nix build
+      - name: Upload om static binary
+        if: matrix.system != 'x86_64-darwin'
+        uses: actions/upload-artifact@v4
+        with:
+          name: om-${{ matrix.system }}
+          path: ./result/bin/om
```
srid, line 48: Let's use `nix build --no-link --print-out-path` and then pass that path between github steps (see [here](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable)).

[Explicit is better than implicit.](https://peps.python.org/pep-0020/)

nix/rust.nix
```diff
@@ -39,7 +39,8 @@
                   Security
                   SystemConfiguration
                 ] ++ [
-                  libiconv
+                  # error: don't yet have a `targetPackages.darwin.LibsystemCross for x86_64-apple-darwin`
```
srid, line 42: Can you elaborate on this comment? Is this an actual error message printed on Intel Mac?

srid (commented):
@shivaraj-bh How should I test the binary on macOS? Currently getting:

image

shivaraj-bh:
> @shivaraj-bh How should I test the binary on macOS? Currently getting:
> 
> image

I would assume, just copying and running the binary (after providing execute permissions using `chmod`), like in CI.

nix/rust.nix
shivaraj-bh: Yes, its an actual error message. I will add a short description above the error message.

shivaraj-bh: https://github.com/juspay/omnix/pull/158/commits/3593d1c488265f66a58e3d97c9b447b62c4ecdd1

.github/workflows/ci.yaml
shivaraj-bh: Done

shivaraj-bh:
I believe it’s complete now. In another PR we should also look at releasing a tarball of these static binaries and provide a curl that users can use to download and run `om`.

srid:
> In another PR we should also look at releasing a tarball of these static binaries and provide a curl that users can use to download and run `om`.

Yup.

> I would assume, just copying and running the binary (after providing execute permissions using `chmod`), like in CI.

That's what I did. Anyway, let's deal with it (the user-level usage instructions) in the aforementioned another PR.

srid (approved):

shivaraj-bh:
> @shivaraj-bh How should I test the binary on macOS? Currently getting:
> 
> image

I was able to reproduce this. This happens when you download artifacts from a browser or any other app and try to run it. As in such a case Apple requires the binary to be notarized by them. But this is not a problem when you get the binary using curl.
You can try it:
```sh
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <YOUR-PAT-TOKEN>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/juspay/omnix/actions/artifacts/1780161532/zip --output om.zip
unzip om.zip
./om
```

PR #165, opened 2024-07-18
`om show` nixosConfigurations

shivaraj-bh:


shivaraj-bh:
`om show github:srid/nixos-config`:

```sh
🐧 NixOS Configurations

Run: nixos-rebuild switch --flake .#<name>
───────────┬─────────────
 name      │ description
═══════════┼═════════════
 immediacy │ N/A
───────────┴─────────────
```

(You are viewing the tail of the output)

srid (approved):
LGTM. Could you update the `nix_rs` ChangeLog accordingly?

See https://github.com/juspay/omnix?tab=readme-ov-file#contributing

shivaraj-bh:
> LGTM. Could you update the `nix_rs` ChangeLog accordingly?
> 
> See https://github.com/juspay/omnix?tab=readme-ov-file#contributing

Done

srid (changes requested):
This just occured to me. Could you make the necessary change in `omnix-gui` as well? I don't see `NixOS Configurations' section in the GUI:

image

While we may not care about GUI parity yet (since CLI is the priority), we should at least try to keep it in sync unless the change will be complex.

shivaraj-bh:
image

PR #174, opened 2024-07-24
cli: add shell completion

### LINKED ISSUE #56: `omnix-cli`: Add autocompletion
Using `clap_complete` crate. The nix package should install the completion files.

For reference implementation, see https://github.com/DeterminateSystems/fh/pull/26

---

Autocompletion looks like this (screenshot of `ripgrep` program, written also in Rust):

image

rsrohitsingh682:
Resolves #56


crates/omnix-cli/Cargo.toml
```diff
@@ -21,6 +21,7 @@ omnix = { workspace = true }
 tabled = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
+clap_complete = "4.5.0"
```
srid, line 24: This should be a workspace dependency, like the rest.

crates/omnix-cli/src/command/mod.rs
```diff
@@ -1,10 +1,16 @@
...
 pub mod ci;
 pub mod health;
 pub mod init;
 pub mod show;
 
+use anyhow::Ok;
+use clap::CommandFactory;
+use clap_complete::generate;
+use std::io;
+
```
srid, line 13: Do not spread out `use` imports around `mod` declarations.

crates/omnix-cli/src/command/mod.rs
```diff
@@ -1,10 +1,16 @@
...
+use crate::Args;
 pub mod ci;
 pub mod health;
 pub mod init;
 pub mod show;
 
+use anyhow::Ok;
+use clap::CommandFactory;
+use clap_complete::generate;
+use std::io;
```
srid, line 12: Prefer to use `std::io` qualified inline.

crates/omnix-cli/src/command/mod.rs
```diff
@@ -1,10 +1,16 @@
 use clap::Subcommand;
 
+use crate::Args;
 pub mod ci;
 pub mod health;
 pub mod init;
 pub mod show;
 
+use anyhow::Ok;
```
srid, line 9: I don't think you need this?

crates/omnix-cli/src/command/mod.rs
```diff
@@ -23,6 +35,11 @@ impl Command {
             Command::Init(config) => config.run().await,
             Command::CI(config) => config.run().await,
             Command::Health(config) => config.run().await,
+            Command::Completion { shell } => {
+                let mut cli = Args::command();
+                generate(*shell, &mut cli, "om", &mut io::stdout());
+                Ok(())
+            }
```
srid, line 42: Move all of this to `command/completion.rs` to be consistent with the rest of the subcommands.

srid (changes requested):
LGTM, but needs final refactor.

srid (approved):

PR #183, opened 2024-07-27
fix: Intel build and add to CI

rsrohitsingh682:
omnix should build on intel-mac (x86_64-darwin) and can be added to ci as well.


rsrohitsingh682:
For building on x86_64-darwin we need apple_sdk_11_0 because x86_64-darwin uses the 10.12 SDK by default.
Ref: https://github.com/NixOS/nixpkgs/blob/e97c8b67e3acca3133cce767c36a795dbfda3a99/doc/stdenv/platform-notes.chapter.md?plain=1#L64C3-L64C113

PR #187, opened 2024-07-28
ci: Pin CLI and nix-health on cachix

srid:
So they don't get GC'ed

https://docs.cachix.org/pins


srid:
Fixed some things in `main`. Now it works:

image

image

PR #189, opened 2024-07-30
feat(omnix-cli): add path completion

sinrohit:
Add path completion to omnix-cli

### DIFF
--- a/crates/nixci/src/cli.rs
+++ b/crates/nixci/src/cli.rs
@@ -18,7 +18,6 @@ use crate::{
     },
 };
 
-
 /// A reference to some flake living somewhere
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum FlakeRef {


crates/nix_rs/src/flake/url.rs
```diff
@@ -9,14 +9,21 @@ use std::{
...
 ///
 /// Use `FromStr` to parse a string into a `FlakeUrl`. Or `From` or `Into` if
 /// you know the URL is valid.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-pub struct FlakeUrl(pub String);
+#[cfg_attr(feature = "clap", derive(Parser))]
+pub struct FlakeUrl {
+    #[cfg_attr(feature = "clap", clap(value_hint = ValueHint::AnyPath))]
+    pub url: String,
+}
```
srid, line 26: There should be a way to obviate this change -- i.e., add value_hint for the whole structure. That or, you could just add cfg_attr to existing `pub String` attr.

rsrohitsingh682: > add value_hint for the whole structure.
AFAIK clap does not allow this way.

Anyways these changes are not required for it to work, we have to annotate FlakeUrl where we want path completions. 
Like this 
Screenshot 2024-08-01 at 6 55 49 PM

srid: Sure, let's do the minimal change required to get this merged. We can refactor later.

srid:
Discussed with @rsrohitsingh682. Closing, because there's no cleaner way to do this, yet. 

Perhaps ask in https://github.com/clap-rs/clap/discussions as to how we can improve `clap` itself to make this simple.

PR #212, opened 2024-08-08
feat: Remote CI builds

### LINKED ISSUE #188: Add `om ci build --on`
Basically https://github.com/srid/nixos-config/blob/master/packages/om-ci-build-remote.nix but upstreamed here. 

The idea is to have `om ci build --on ssh://user@host` in effect build the flake on the remote `host`, rather than locally (as it will by default).

rsrohitsingh682:
Resolves  #188 

You can run om ci like:

`om ci build --on user@ip-address`

Using nix

`nix run github:juspay/omnix/om-ci-remote-build -- ci run --on user@ip`


crates/nix_rs/src/flake/metadata.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct FlakeMetadata {
+    pub path: String,
+}
+
+/// Runs `nix flake metadata json` in Rust
+pub async fn get_flake_metadata_json(
```
srid, line 11: This should following the naming convention (see `config.rs`, `outputs.rs` for example).

crates/nix_rs/src/flake/metadata.rs
```diff
@@ -0,0 +1,19 @@
+use crate::command::{NixCmd, NixCmdError};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct FlakeMetadata {
+    pub path: String,
```
srid, line 7: We should try not use `str`/ `String` for everything. When there's an appropriate type -- and there's one for paths -- we should try to use that.

crates/nixci/crate.nix
```diff
@@ -27,6 +27,7 @@ in
         pkgs.openssl
       ];
       DEVOUR_FLAKE = inputs.devour-flake;
+      OMNIX = inputs.omnix;
```
srid, line 30:
```suggestion
      OMNIX_SOURCE = inputs.omnix;
```

crates/nixci/src/lib.rs
```diff
@@ -55,6 +60,29 @@ pub async fn nixci(
     }
 }
 
+async fn remote_build(
+    cmd: &NixCmd,
+    build_cfg: &BuildConfig,
+    cfg: &config::Config,
+    host: &String,
+) -> anyhow::Result<Vec<StorePath>> {
+    let omnix_input = format!("{}", env!("OMNIX"));
```
srid, line 69: In consistent with how we use `env!`, make this a global const static.

crates/nixci/src/lib.rs
```diff
@@ -55,6 +60,29 @@ pub async fn nixci(
...
+    let metadata = nix_rs::flake::metadata::get_flake_metadata_json(cmd, &cfg.flake_url.0).await?;
+
+    nix_rs::copy::run_nix_copy(
+        cmd,
+        &host,
+        &omnix_input,
+        &metadata.path,
+        &build_cfg.extra_nix_build_args,
+    )
+    .await?;
```
srid, line 80: You need to `nix copy` the paths in `--override-inputs` too (if specified).

srid: Similar to:

https://github.com/srid/nixos-flake/blob/5734c1d9a5fe0bc8e8beaf389ad6227392ca0108/activate/activate.nu#L103-L107

flake.nix
```diff
@@ -25,6 +25,9 @@
     # Also note: Using shivaraj-bh fork of nix which fixes x86_64-darwin on top of github:DeterminateSystems/nix-src/flake-schemas
     nix.url = "github:shivaraj-bh/nix/flake-schemas";
 
+    omnix.url = "github:juspay/omnix";
+    omnix.flake = false;
```
srid, line 29: This is wrong. Use `inputs.self`

nix/modules/devshell.nix
```diff
@@ -38,6 +38,7 @@ in
         ''
           # For nixci
           export DEVOUR_FLAKE=${inputs.devour-flake}
+          export OMNIX=${inputs.omnix}
```
srid, line 41: This should be in around line 20 above.

I'll move DEVOUR_FLAKE there as well later.

crates/nixci/src/lib.rs
rsrohitsingh682: I handled it in this way.
(https://github.com/juspay/omnix/pull/212/files#diff-356b04180edb27abcd766667939d7a544466fdc67df4005f041453532a9a87e8R16-R29)

srid:
We should unit tests as well. That the correct SSH CLI is constructed. We don't have to run `nix copy` in the tests.

crates/nixci/src/lib.rs
srid: It shouldn't be done that way. See this link for why:

https://en.wikipedia.org/wiki/Separation_of_concerns

srid: Also, the logic is wrong.

srid: Because, we are supposed to run `nix copy` **for each** entry in `--overrride-inputs` list.

srid: > Also, the logic is wrong.

I could be wrong, though (is `nix copy` automagically copying those paths?). Need to test to know for sure.

crates/nix_rs/src/copy.rs
```diff
@@ -0,0 +1,37 @@
+use crate::command::{CommandError, NixCmd};
+
+/// Runs `nix copy` in Rust
+pub async fn run_nix_copy(
+    cmd: &NixCmd,
+    host: &str,
+    omnix_input: &str,
+    path: &str,
+    extra_args: &[String],
```
srid, line 9: Per https://en.wikipedia.org/wiki/Separation_of_concerns you would delete these arguments and put a singular `Vec<PathBuf>` argument in their place.

crates/nixci/src/lib.rs
srid: Interesting, looks like it *does* automagically copy it? 

image

@rsrohitsingh682 Can you make it so that the stdout of `nix copy` is shown in omnix console output as well?

crates/nix_rs/src/copy.rs
srid: If this <https://github.com/juspay/omnix/pull/212#discussion_r1711652998> is confirmed, we could take additional `nix_args` as well, and the caller would only pass `--override-inputs` in it.

srid:
I'm doing some refactor on `main`. Don't worry about rebasing just yet (until this PR is ready and reviewed).

srid:
@rsrohitsingh682 `main` has been refactored. You can build on top of it. It probably is easier to just reapply your changes on top. You can start from 

https://github.com/juspay/omnix/blob/main/crates/nixci/src/command/build.rs

srid:
Heads up: `om ci build` is now `om ci run`.

You can find the command entry point in,
https://github.com/juspay/omnix/blob/main/crates/nixci/src/command/run.rs

rsrohitsingh682:
CI failed because docs are not present for the files that I added. Will add those tomorrow.

srid:
I need to fix the doc targets, but meanwhile you can use:

```
cargo-watch -x cargo doc -p nixci
```

srid:
(Zed or VSCode should also show these as warnings)

crates/nix_rs/src/copy.rs
```diff
@@ -0,0 +1,19 @@
+use crate::command::{CommandError, NixCmd};
+use std::path::PathBuf;
+
+/// Runs `nix copy` in Rust
+pub async fn from_nix(cmd: &NixCmd, host: &str, paths: Vec<&PathBuf>) -> Result<(), CommandError> {
```
srid, line 5: Prefer `&[PathBuf]` over mandating the user to build `Vec` every time.

crates/nix_rs/src/copy.rs
```diff
@@ -0,0 +1,19 @@
+use crate::command::{CommandError, NixCmd};
+use std::path::PathBuf;
+
+/// Runs `nix copy` in Rust
```
srid, line 4: This should be normal function doc without relying on user's knowledge of `nix copy` (which command should be mentioned as sidenote, however).

srid: The same applies for other parts of this PR.

crates/nix_rs/src/flake/metadata.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use super::url::FlakeUrl;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct FlakeMetadata {
+    pub path: PathBuf,
+}
+
+/// Runs `nix flake metadata json` in Rust
+pub async fn from_nix(cmd: &NixCmd, flake_url: &FlakeUrl) -> Result<FlakeMetadata, NixCmdError> {
```
srid, line 14: Be consistent when it comes to writing API code.

In the rest of the project, is `from_nix` in fact a top-level function?

crates/nix_rs/src/flake/metadata.rs
```diff
@@ -0,0 +1,24 @@
...
+    let json = cmd
```
srid, line 15:
```suggestion
    let metadata = cmd
```

Obviously this value is not a JSON string.

crates/nixci/src/command/core.rs
```diff
@@ -39,7 +39,10 @@ impl Command {
         tracing::info!("{}", "\n👟 Reading om.ci config from flake".bold());
         let cfg = self.get_config(nixcmd).await?;
         match self {
-            Command::Run(cmd) => cmd.run(nixcmd, verbose, cfg).await,
+            Command::Run(cmd) => match &cmd.steps_args.build_step_args.on {
+                Some(host) => cmd.run_remote(nixcmd, cfg, host).await,
+                None => cmd.run(nixcmd, verbose, cfg).await,
+            },
```
srid, line 45: The `run` command is responsible for ... well ... all that `om ci run` does. Ergo, remote run logic should not spill over here; rather it should remain encapsulated in `command/run.rs` (you can still keep that `run_remote` function but as private function to that module).

crates/nixci/src/command/run.rs
```diff
@@ -9,13 +9,21 @@ use nix_rs::{
...
     config::core::Config,
     flake_ref::FlakeRef,
-    nix::system_list::{SystemsList, SystemsListFlakeRef},
+    nix::{
+        ssh,
+        system_list::{SystemsList, SystemsListFlakeRef},
+    },
 };
 
+/// We expect this environment to be set in Nix build and shell.
```
srid, line 23: Generally, always document the "what" (rather than "why" or any other thing).

```suggestion
/// Path to Rust source corresponding to this (running) instance of Omnix
```

crates/nixci/src/command/run.rs
```diff
@@ -9,13 +9,21 @@ use nix_rs::{
...
+pub const OMNIX_SOURCE: &str = env!("OMNIX_SOURCE");
```
srid, line 24: Cannot this be made of type `PathBuf` (to obviate the conversion below where it gets used) here?

crates/nixci/src/command/run.rs
```diff
@@ -76,6 +84,28 @@ impl RunCommand {
         Ok(())
     }
 
+    /// Run the ci run command on remote
+    pub async fn run_remote(&self, nixcmd: &NixCmd, cfg: Config, host: &str) -> anyhow::Result<()> {
+        let metadata = nix_rs::flake::metadata::from_nix(nixcmd, &cfg.ref_.flake_url).await?;
+
+        let omnix_source = PathBuf::from(OMNIX_SOURCE);
+
+        let args = vec![&omnix_source, &metadata.path];
```
srid, line 93: You can inline `args` where it gets used, immediately below.

crates/nix_rs/src/copy.rs
srid: Also, `from_nix` is only used when reading a certain Rust type from Nix. `nix copy` on the other hand does no reading. This should be called `nix_copy`

crates/nixci/src/nix/ssh.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// Runs `nix run command through ssh on remote machine` in Rust
+///
+pub async fn on_ssh(
+    build_step_args: &BuildStepArgs,
+    remote_address: &str,
+    omnix_input: &Path,
+    flake_url: PathBuf,
+    cfg_ref: ConfigRef,
+) -> anyhow::Result<()> {
```
srid, line 17: @rsrohitsingh682 I think you need to rethink this function entirely. 

First of all, anything under the `nix/` package should *not* be depending on the rest of the `nixci` crate. Which is kind of obvious.  The `nix/` package contains Nix-specific stuff; not nixci specific stuff. The use of both `BuildStepArgs` and `ConfigRef` makes this otherwise general interface of the module **complect** with the rest of nixci. See https://www.infoq.com/presentations/Simple-Made-Easy/

For that similar reason, this module should not of course be aware of the application that uses it; i.e, Omnix. Ergo, there should not be an argument called `omnix_input`. 

You should also document the function in full, and explain what each argument does.

srid: Naturally, the body of the function too complects, thus violating simplicity principle (per Rich Hickey presentation above).

srid (changes requested):
I think the PR behaves as expected, though the output (where the remote switch over happens) can be made nicer.

Main thing left to do is refactor.

crates/nixci/src/command/run.rs
rsrohitsingh682: it's not possible to directly make it a const PathBuf because  `env! macro` returns a &'static str, which is a string slice. Also creating a `PathBuf` requires heap allocation, which can't be done at compile-time in a const context.

crates/nixci/src/command/run.rs
```diff
@@ -76,6 +92,70 @@ impl RunCommand {
...
+        .collect();
+
+        // call ci run on remote machine through ssh
+        ssh::on_ssh(host, &nix_run_args).await?;
+
+        Ok(())
+    }
+
+    // Return ci run args along with build_step_args
+    fn get_ci_run_args_for_remote(
```
rsrohitsingh682, line 122: I am still doubting that this would be the correct way to go about it. Little advice would be appreciated.

srid:
Alright, cheers. Let me take over from here.

PR #224, opened 2024-08-19
ci: Avoid redundant nix-store invocations when `--print-all-dependencies`

### LINKED ISSUE #220: `--print-all-dependencies` is running nix-store multiple times
## Currently

image

## Ideally

All these paths can be passed to the _same_ `nix-store` invocation:

image

rsrohitsingh682:
Resolves #220

### DIFF
--- a/crates/nix_rs/src/store.rs
+++ b/crates/nix_rs/src/store.rs
@@ -70,11 +70,9 @@ impl NixStoreCmd {
             let drv = self.nix_store_query_deriver(out.clone()).await?;
             all_drvs.insert(drv);
         }
-        let mut all_outs = HashSet::new();
-        for drv in all_drvs {
-            let deps = self.nix_store_query_requisites_with_outputs(drv).await?;
-            all_outs.extend(deps.into_iter());
-        }
+        let all_outs = self
+            .nix_store_query_requisites_with_outputs(all_drvs)
+            .await?;
         Ok(all_outs)
     }
 
@@ -106,19 +104,20 @@ impl NixStoreCmd {
         }
     }
 
-    /// Given a derivation path, this function recursively queries and return all
+    /// Given the derivation paths, this function recursively queries and return all
     /// of its dependencies in the Nix store.
     pub async fn nix_store_query_requisites_with_outputs(
         &self,
-        drv_path: PathBuf,
-    ) -> Result<Vec<StorePath>, NixStoreCmdError> {
+        drv_paths: HashSet<PathBuf>,
+    ) -> Result<HashSet<StorePath>, NixStoreCmdError> {
         let mut cmd = self.command();
-        cmd.args([
-            "--query",
-            "--requisites",
-            "--include-outputs",
-            drv_path.to_string_lossy().as_ref(),
-        ]);
+        let mut args = vec!["--query", "--requisites", "--include-outputs"];
+        let paths: Vec<&str> = drv_paths
+            .iter()
+            .map(|path| path.as_path().to_str().unwrap_or_default())
+            .collect();
+        args.extend(paths);
+        cmd.args(args);
         crate::command::trace_cmd(&cmd);
         let out = cmd.output().await?;
         if out.status.success() {
--- a/crates/nix_rs/src/store.rs
+++ b/crates/nix_rs/src/store.rs
@@ -166,14 +166,17 @@ impl NixStoreCmd {
             .args(out_paths.iter().map(StorePath::as_path));
 
         crate::command::trace_cmd(&cmd);
+
         let out = cmd.output().await?;
         if out.status.success() {
-            let out = String::from_utf8(out.stdout)?.trim().to_string();
-            let drv_paths: HashSet<PathBuf> = out.lines().map(PathBuf::from).collect();
+            let drv_paths: Vec<PathBuf> = String::from_utf8(out.stdout)?
+                .lines()
+                .map(PathBuf::from)
+                .collect();
             if drv_paths.contains(&PathBuf::from("unknown-deriver")) {
                 return Err(NixStoreCmdError::UnknownDeriver);
             }
-            Ok(drv_paths.into_iter().collect())
+            Ok(drv_paths)
         } else {
             // TODO(refactor): When upstreaming this module to nix-rs, create a
             // nicer and unified way to create `ProcessFailed`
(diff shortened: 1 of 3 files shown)


crates/nix_rs/src/store.rs
```diff
@@ -106,19 +104,20 @@ impl NixStoreCmd {
...
     }
 
-    /// Given a derivation path, this function recursively queries and return all
+    /// Given the derivation paths, this function recursively queries and return all
     /// of its dependencies in the Nix store.
     pub async fn nix_store_query_requisites_with_outputs(
         &self,
-        drv_path: PathBuf,
-    ) -> Result<Vec<StorePath>, NixStoreCmdError> {
+        drv_paths: HashSet<PathBuf>,
```
srid, line 111: This should take a slice (`&[Pathbuf]`). There's not reason for it to require a HashSet. A convenience of a single caller should not dictate API of some core function.

crates/nix_rs/src/store.rs
```diff
@@ -106,19 +104,20 @@ impl NixStoreCmd {
...
-        cmd.args([
-            "--query",
-            "--requisites",
-            "--include-outputs",
-            drv_path.to_string_lossy().as_ref(),
-        ]);
+        let mut args = vec!["--query", "--requisites", "--include-outputs"];
+        let paths: Vec<&str> = drv_paths
+            .iter()
+            .map(|path| path.as_path().to_str().unwrap_or_default())
```
srid, line 117: @rsrohitsingh682 Is this code behaviourally correct? What do you think happens in the default (`unwrap_or_default`) case?

srid: Also, you want to understand how the `args` method works.  What does `AsRef` do? How do you pass it a slice of `PathBuf`?

srid: I see the same code in #212. We should not reflexively use Default instance without understanding what it does.

rsrohitsingh682:
Converting this to draft as I would like to do similar changes to `nix_store_query_requisites_with_outputs` function as well.

crates/nix_rs/src/store.rs
```diff
@@ -147,38 +147,41 @@ impl NixStoreCmd {
...
         crate::command::trace_cmd(&cmd);
         let out = cmd.output().await?;
         if out.status.success() {
-            let drv_path = String::from_utf8(out.stdout)?.trim().to_string();
-            if drv_path == "unknown-deriver" {
+            let out = String::from_utf8(out.stdout)?.trim().to_string();
+            let drv_paths: HashSet<PathBuf> = out
+                .lines()
+                .map(|l| l.trim().to_string())
+                .filter(|l| !l.is_empty())
```
srid, line 178: Why do we need these two lines? What happens if you get rid of them?

crates/nix_rs/src/store.rs
```diff
@@ -188,19 +191,15 @@ impl NixStoreCmd {
...
 
-    /// Given a derivation path, this function recursively queries and return all
+    /// Given the derivation paths, this function recursively queries and return all
     /// of its dependencies in the Nix store.
     pub async fn nix_store_query_requisites_with_outputs(
         &self,
-        drv_path: PathBuf,
-    ) -> Result<Vec<StorePath>, NixStoreCmdError> {
+        drv_paths: &[PathBuf],
+    ) -> Result<HashSet<StorePath>, NixStoreCmdError> {
```
srid, line 199: This should continue to return `Vec` (so as to maintain original order).

Also, I see no reason for it to require a HashSet. The convenience of a single caller should not universally dictate API of some core function.

crates/nix_rs/src/store.rs
```diff
@@ -147,38 +147,41 @@ impl NixStoreCmd {
...
     }
 
-    /// Return the derivation used to build the given build output.
+    /// Return the derivations used to build the given build output.
     pub async fn nix_store_query_deriver(
         &self,
-        out_path: StorePath,
-    ) -> Result<PathBuf, NixStoreCmdError> {
+        out_paths: &[StorePath],
+    ) -> Result<HashSet<PathBuf>, NixStoreCmdError> {
```
srid, line 163: This should return `Vec` (so as to maintain original order).

Also, I see no reason for it to require a HashSet over Vec. The convenience of a single caller should not universally dictate API of some core function.

crates/nix_rs/src/store.rs
```diff
@@ -147,38 +147,41 @@ impl NixStoreCmd {
...
-        cmd.args([
-            "--query",
-            "--valid-derivers",
-            out_path.as_path().to_string_lossy().as_ref(),
-        ]);
+        let mut paths = vec![];
+        for path in out_paths {
+            paths.push(path.as_path());
+        }
+        cmd.args(["--query", "--valid-derivers"]).args(paths);
```
srid, line 169: Cannot this be simplified as:

```rust
cmd.args(["--query", "--valid-derivers"])
   .args(out_paths.iter().map(StorePath::as_path));
```

crates/nix_rs/src/store.rs
rsrohitsingh682: Even I thought about this but then chose HashSet thinking about consistency as main function `fetch_all_deps` also returns a HashSet and the other reason was HashSet would contain unique while Vec won't.

rsrohitsingh682: But thinking about this `The convenience of a single caller should not universally dictate API of some core function.` now I am thinking I can collect the final result in a HashSet and return.

crates/nix_rs/src/store.rs
rsrohitsingh682: You're right. These are actually not needed as we have applied trim() before to the whole String so chances of getting whitespace or empty lines won't be there. I can safely remove these two extra lines.

rsrohitsingh682: I tested it, it behaves normally.

crates/nix_rs/src/store.rs
```diff
@@ -147,38 +147,33 @@ impl NixStoreCmd {
...
+        cmd.args(["--query", "--valid-derivers"])
+            .args(out_paths.iter().map(StorePath::as_path));
+
         crate::command::trace_cmd(&cmd);
         let out = cmd.output().await?;
         if out.status.success() {
-            let drv_path = String::from_utf8(out.stdout)?.trim().to_string();
-            if drv_path == "unknown-deriver" {
+            let out = String::from_utf8(out.stdout)?.trim().to_string();
+            let drv_paths: HashSet<PathBuf> = out.lines().map(PathBuf::from).collect();
```
rsrohitsingh682, line 172: Converting it to HashSet first and later to Vec while returning  as I want this check for `unknown-deriver` in constant time.

srid (approved):
Good, for future - remember to update CHANGELOG and release history.md

Issue #216, opened 2024-08-12
`om ci`: Configurable steps


srid:
Create a notion of a "[CI] step".

Make the current devour-flake build a builtin step, called `build`, as well as the main Nix version check a builtin step, called `health`. Add additional builtin checks, like `cachix` (solving #197). Finally, allow users to define their own steps, like `closure-size-check` and `cli-tests` (solving #198). We can even solve #200 by writing a builtin step. 

The user can enable or disable any step, including builtin ones.

The config can be in the same place. Note the use of string interpolation to refer to store paths in cachix pin configuration, since we can't use functions or modules here due to JSON conversion requirement.

<img width="496" alt="image" src="https://github.com/user-attachments/assets/6da5848f-4380-4a95-b71f-949558d52780">

srid:
We don't need to rely on flake apps -- thus there's no need for "CI apps" -- since the `command` option can run about any command.

srid:
This is by and large done: https://omnix.page/om/ci.html#custom

PR #229, opened 2024-08-21
`om ci`: Run `nix flake check` after build

### LINKED ISSUE #200: `om ci`: Run `nix flake check`
Not all checks are done by building a derivation (what `nixci build` does currently).

For eg., `nix flake check` tests for unknown properties in `templates.*` and errors out:

https://github.com/NixOS/nix/blob/babfd0cd99012df9d645d38e47fa31d049bf01d8/src/nix/flake.cc#L550-L554

image

The above will not be caught by `nixci build`. 

Do we care to support this niche case, though?

--

More generally, `nix flake check` does a bunch of "Evaluation checks":

image

srid:
Add a new CI step (cf. #216) to `nixci` so that `nix flake check` is run for all sub-flakes as part of `om ci run`.

Resolves #200 

image

### DIFF
--- a/crates/nixci/CHANGELOG.md
+++ b/crates/nixci/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - New
   - Introduced notion of 'steps'. Renamed 'build' to 'run'.
+    - Added a step to run `nix flake check`
 - `config.rs`: Refactored to change API.
 
 ## 1.1.0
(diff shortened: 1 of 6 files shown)


srid:
This caught an issue with services-flake:

https://github.com/juspay/omnix/actions/runs/10493657389/job/29068127060

image

PR #222, opened 2024-08-16
nix_rs: Inherit `stderr` while running `nix` as a child process

shivaraj-bh:
resolves #209 

We also use `flake-schemas` in this PR to fetch the `om` config from the flake output instead of using `nix eval` (since inheriting `stderr` means it's no longer captured for `om` to process). An additional, unintended benefit of using `flake-schemas` is caching the output of `nix flake show`.

## Changes in nix CLI
https://github.com/shivaraj-bh/nix/commit/1d23c1e871981f5666a12c4409bd0574fc1e1e02

## Demo

```sh
❯ om show github:nammayatri/nammayatri
🐚 nix --extra-experimental-features 'nix-command flakes' show-config --json
warning: unknown setting 'upgrade-nix-store-path-url'
🐚 /nix/store/n02w2ybg9fc78grzz9i2aj49q3rysp7m-nix-2.24.0pre20240801_af10904/bin/nix flake show --legacy --allow-import-from-derivation --json --default-flake-schemas /nix/store/xzalq6mcw0ahyaccab6k98dbx3ll53y6-source github:nammayatri/nammayatri
warning: unknown experimental feature 'repl-flake'
[0.6 MiB DL] downloading 'https://github.com/nammayatri/nammayatri/archive/a3e472efe24e731268d20a31f75a1519ded50220.tar.gz'^C
[0.6 MiB DL]
```

## TODO
- [x] Update `nix_rs` changelog
- [x] Update `history.md`
- [x] `flake-schemas` as local flake (for `omSchema`) while importing majority of schemas from DetSys/flake-schemas
- [x] move `find_qualified_attr_in_flake_output` to `FlakeOutputs`
- [x] simplify `FlakeOutputs` serialization… use untagged for flattening
- [x] refactor(DRY) in `nix/flake-schemas/flake.nix`

PR #235, opened 2024-08-26
nix_rs: Don't hardcode flake schema types

### LINKED ISSUE #262: Don't hardcode flake schema types
https://github.com/juspay/omnix/blob/c566f7a4566473fdedca6bb8d32178820be0502f/crates/nix_rs/src/flake/schema.rs#L16-L35
To avoid maintaining these types in rust

shivaraj-bh:
resolves #262

TODO:
- [x] Everything about flake-schemas must be in schema.rs


shivaraj-bh:
I will also have to re-think this design based on how it will be used in https://github.com/juspay/omnix/pull/222, to parse `om` flake output.

crates/nix_rs/src/flake/outputs.rs
```diff
@@ -1,109 +1,43 @@
...
-            Self::Leaf(v) => Some(v),
+            Self::Val(v) => Some(v),
             _ => None,
         }
     }
 
-    /// Ensure the value is an attrset, and get it
-    pub fn as_attrset(&self) -> Option<&BTreeMap<String, InventoryItem>> {
+    /// Get the attrset as a vector of key-value pairs
+    pub fn as_vec(self) -> Vec<(String, Val)> {
```
srid, line 26: How does this function compare to implementing an `Iterator` trait?

crates/nix_rs/src/flake/outputs.rs
```diff
@@ -1,109 +1,43 @@
...
         match self {
-            Self::Leaf(v) => Some(v),
+            Self::Val(v) => Some(v),
             _ => None,
         }
     }
 
-    /// Ensure the value is an attrset, and get it
-    pub fn as_attrset(&self) -> Option<&BTreeMap<String, InventoryItem>> {
+    /// Get the attrset as a vector of key-value pairs
```
srid, line 25: Does this description capture the whole semantics?

I had to read the logic of implementation to understand this basically returns the *immediate* children, but only `Val`.

srid: Perhaps, then, this should be called `get_children`?

crates/nix_rs/src/flake/outputs.rs
```diff
@@ -1,109 +1,43 @@
...
-        let v = nix_eval::<Self>(nix_cmd, &flake_opts, &inspect_flake).await?;
-        Ok(v)
-    }
-}
-
-impl InventoryItem {
-    /// Get the non-attrset leaf
-    pub fn as_leaf(&self) -> Option<&Leaf> {
+    /// Get the non-attrset value
+    pub fn as_val(&self) -> Option<&Val> {
```
srid, line 18: Likewise, `get_value`?

(API naming is generally best reflective of semantics than Rust types)

crates/nix_rs/src/flake/outputs.rs
srid: Maybe not, based on how I understand the semantics of this function to be.

crates/nix_rs/src/flake/schema.rs
```diff
@@ -1,108 +1,216 @@
...
+  /// We expect this environment to be set in Nix build and shell.
+  pub static ref INSPECT_FLAKE: FlakeUrl = {
+    Into::<FlakeUrl>::into(Path::new(env!("INSPECT_FLAKE")))
+  };
+}
+
+/// Represents the schema of a given flake evaluated using [static@INSPECT_FLAKE]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FlakeSchemas {
+    pub inventory: BTreeMap<String, InventoryItem>,
```
srid, line 30: Since Nix attrsets have no concept of order, shouldn't we just use `HashMap` here (and in `outputs.rs`)?

crates/nix_rs/src/flake/schema.rs
```diff
@@ -1,108 +1,216 @@
...
-            if let Some(item) = inventory_item {
-                if let Some(out) = item.pop(ks) {
-                    if let Some(outs) = out.as_attrset() {
-                        for (k, v) in outs {
-                            if let Some(leaf) = v.as_leaf() {
-                                result.insert(k.clone(), leaf.clone());
-                            }
-                        }
+impl FlakeSchemas {
+    /// Determine flake outputs using [static@INSPECT_FLAKE] and [static@DEFAULT_FLAKE_SCHEMAS]
```
srid, line 41:
```suggestion
    /// Get the [FlakeSchema] for the given flake
    /// 
    /// This uses [static@INSPECT_FLAKE] and [static@DEFAULT_FLAKE_SCHEMAS]
```

crates/nix_rs/src/flake/schema.rs
```diff
@@ -1,108 +1,216 @@
...
-                &mut output.inventory.get_mut("homeConfigurations"),
-                &["children"],
-            ),
-            nixos_modules: pop_tree(&mut output.inventory.get_mut("nixosModules"), &["children"]),
-            docker_images: pop_tree(&mut output.inventory.get_mut("dockerImages"), &["children"]),
-            overlays: pop_tree(&mut output.inventory.get_mut("overlays"), &["children"]),
-            templates: pop_tree(&mut output.inventory.get_mut("templates"), &["children"]),
-            schemas: pop_tree(&mut output.inventory.get_mut("schemas"), &["children"]),
-            other: Some(output.clone()),
+/// The metadata of a flake output value which is of non-attrset [Type]
```
srid, line 125: Might as well plainly call it "terminal value" (no children) instead of leverage understanding of Rust types.

srid (commented):
Just reviewed. @shivaraj-bh Can you merge from `main` to bring in #283 and document the new types/ functions accordingly?

(Bear in mind that documentation should be more about general functional semantics than verbatim repetition of types).

crates/nix_rs/src/flake/schema.rs
```diff
@@ -1,109 +1,200 @@
...
+    DevShell,
+    #[serde(rename = "CI test")]
+    Check,
+    #[serde(rename = "app")]
+    App,
+    #[serde(rename = "template")]
+    Template,
+    #[serde(other)]
+    Unknown,
+}
```
shivaraj-bh, line 177: @srid I have documented everything, except for this enum. Do these require documentation as well?

crates/nix_rs/src/flake/schema.rs
shivaraj-bh: Done in https://github.com/juspay/omnix/pull/235/commits/9de27ef22bc18001c82b358cd3bc37ed64016e59

crates/nix_rs/src/flake/schema.rs
srid: `#[allow(missing_docs)]` looks like a legitimate use here; since these variants are self-documenting anyway.

crates/nix_rs/src/flake/outputs.rs
shivaraj-bh: https://github.com/juspay/omnix/pull/235/commits/0432202ef32dff1fb74710f09b368b82e47d0380

crates/nix_rs/src/flake/outputs.rs
shivaraj-bh: https://github.com/juspay/omnix/pull/235/commits/c877c77653fb248ece5add7ab7741c8055d3635a

crates/nix_rs/src/flake/schema.rs
shivaraj-bh: https://github.com/juspay/omnix/pull/235/commits/e5c6d6a46b31148d2f44350e9e2275d23eb53036

crates/nix_rs/src/flake/outputs.rs
srid: Unresolving this, because this part is not addressed:


> Does this description capture the whole semantics?
> 
> I had to read the logic of implementation to understand that this function basically returns the _immediate_ children, but only `Val`.

I think you want to describe it as something like "Get the immediate terminal-value children of this [FlakeOutput]".

crates/nix_rs/src/flake/schema.rs
```diff
@@ -1,109 +1,206 @@
...
+    #[serde(rename = "development environment")]
+    DevShell,
+    #[serde(rename = "CI test")]
+    Check,
+    #[serde(rename = "app")]
+    App,
+    #[serde(rename = "template")]
+    Template,
+    #[serde(other)]
+    Unknown,
```
srid, line 182: Perhaps this should be `Other(String)`?  Assuming flake-schemas passes the arbitrary type string, that is. Not sure.

srid: Can address this in different PR. I'm merging this.

srid (approved):
Good stuff 👍🏿

PR #243, opened 2024-08-29
Support Nix 2.20

### LINKED ISSUE #114: Nix 2.20 support
Make sure that nix-browser and its crates work on Nix 2.20,

https://github.com/juspay/nix-browser/blob/0f19029ad82f94a830f0b8f3a67283838f71b672/crates/nix_rs/src/config.rs#L45-L47

image

https://nixos.org/manual/nix/stable/release-notes/rl-2.20.html

rsrohitsingh682:
resolves #114


crates/nix_rs/src/config.rs
```diff
@@ -59,9 +62,21 @@ impl NixConfig {
     pub async fn from_nix(
         nix_cmd: &super::command::NixCmd,
     ) -> Result<NixConfig, super::command::NixCmdError> {
-        let v = nix_cmd
-            .run_with_args_expecting_json(&["show-config", "--json"])
-            .await?;
+        let nix_version = NixVersion::from_nix().await?;
```
rsrohitsingh682, line 65: This doesn't seem like a very good approach to me, still thinking how this can be handled better.

srid: It may well require some refactoring outside of the context of this code. I'll do that.

PR #268, opened 2024-09-12
ci: Cache `github:nix-systems` locally

### LINKED ISSUE #210: `om ci`: Locally cached flakes
This is a general variant of https://github.com/srid/nixci/issues/83

`om ci` already caches `github:srid/devour-flake` (via `DEVOUR_FLAKE` env var). Make it also cache the various `github:nix-systems/*` flakes, so we can use that to substitute `--systems` argument in `om ci`.

srid:
Resolves #210


srid:
image

PR #277, opened 2024-09-15
Use `inspect` to remove the dependency on `nix` fork for flake-schemas

### LINKED ISSUE #261: Remove dependency on `nix` fork
Instead of `nix flake show` from the DetSys nix fork, use https://github.com/DeterminateSystems/inspect

shivaraj-bh:
resolves #261

`inspect` is used to inspect a flake output based on a schema. See: https://github.com/juspay/inspect

See changes on the `juspay/inspect` fork: https://github.com/juspay/inspect/commit/279cfa5c434e5bbebb4c07fa6af2deeb84e03bb0

Todo:

- [x] Write test to verify if `nixosConfigurations` is shown on `darwin` (This is necessary because we evaluate derivations only for the current system to avoid errors due to IFD. However, specific derivations can be skipped if they have `evalOnAllSystems = true;` in their schema, which is the case for outputs like `{nixos, darwin, home}Configurations`. `evalOnAllSystems` is supported in https://github.com/juspay/inspect/commit/279cfa5c434e5bbebb4c07fa6af2deeb84e03bb0. For more context, also see https://github.com/juspay/omnix/pull/277#discussion_r1760164052) **Done in https://github.com/juspay/omnix/pull/277/commits/6303a7deb8ac748234a127c9a7c6cc2ed59bd576**
- [ ] Credit https://github.com/DeterminateSystems/inspect
- [x] @srid Move `SystemsListFlakeRef` to `nix_rs` and use it for the `systems` input of `inspect` flake


crates/nix_rs/src/flake/outputs.rs
```diff
@@ -6,35 +6,69 @@ use std::{
...
+                "--json",
+                "--override-input",
+                "flake-schemas",
+                env!("DEFAULT_FLAKE_SCHEMAS"),
+                "--override-input",
+                "flake",
+                flake_url,
+                "--override-input",
+                "systems",
+                // TODO: Move `SystemsListFlakeRef` to `nix_rs` and use it here.
```
srid, line 53: Let me do this TODO once this PR is ready. I'll have to think about something here.

flake.nix
```diff
@@ -20,11 +20,8 @@
     devour-flake.url = "github:srid/devour-flake";
     devour-flake.flake = false;
 
-    # TODO: Use upstream after https://github.com/NixOS/nix/pull/8892
-    # Note: This version of nix is only used to run `nix flake show` in omnix-cli
-    # Also note: Using shivaraj-bh fork of nix which fixes x86_64-darwin on top of github:DeterminateSystems/nix-src/flake-schemas
-    nix.url = "github:shivaraj-bh/nix/flake-schemas";
-    nix.inputs.flake-parts.follows = "flake-parts";
+    inspect.url = "github:juspay/inspect";
```
srid, line 23: I recommend not pushing to `main` branch on forks.

srid: Also, let's open a PR for this change. Not sure if 1 PR or two PRs, since https://github.com/juspay/inspect/commit/279cfa5c434e5bbebb4c07fa6af2deeb84e03bb0 refers to two things.

nix/flake-schemas/flake.nix
```diff
@@ -33,11 +33,59 @@
               })
             output);
       };
+      nixosConfigurationsSchema = {
```
srid, line 36: These three schemas already exist upstream,
https://github.com/DeterminateSystems/flake-schemas/blob/main/flake.nix

1. Why duplicate them here?
2. If this a fork, we should open PRs upstream.

shivaraj-bh: These exist because of `evalOnAllSystems = true;` config, which is supported on `juspay/inspect`

shivaraj-bh: I will add comments

flake.nix
shivaraj-bh: https://github.com/juspay/omnix/pull/277/commits/2ab059c8ff56596914247485cd711d8588870f5a

shivaraj-bh:
image

@srid `omnix-gui` not disabled in CI?

nix/flake-schemas/flake.nix
shivaraj-bh: Upstream `inspect` by default recursively evaluates the derivations of all the systems supported by a given flake, which can be problematic when IFD is involved. I can’t use `builtins.tryEval` either because IFD errors are not thrown from the nix expressions language.

To avoid errors due to IFD, I introduced `systems` flake input. This means no more showing `nixosConfigurations` on darwin or vice-versa, because the flake outputs are filtered based on `forSystems` attribute in the schema of a given output, which is usually `x86_64-linux` or `aarch64-linux`. This is why `evalOnAllSystems` exists: to bypass `forSystems` based filtering on  a given schema. 

I am not pleased with this approach, let me know if this can be solved better

srid:
> `omnix-gui` not disabled in CI?

It is disabled in Nix; this PR's failure is from devshell's `cargo test`. Does 4be3a71cd819609a9c4ee76862d870d2a8deb6b7 help?

flake.nix
srid: > Also, let's open a PR for this change. Not sure if 1 PR or two PRs, since [juspay/inspect@279cfa5](https://github.com/juspay/inspect/commit/279cfa5c434e5bbebb4c07fa6af2deeb84e03bb0) refers to two things.

Rationale: we want to move away from the culture of maintaining monolithic fork branches for long time. 

Ideally there would be two PRs upstream -- one for specifying flake-schemas as input, and another for adding `evalOnAllSystems` (I imagine the 2nd one will see some discussion among maintainers).

Then, we can have one branch in our fork, where commits from these two PRs are merged - and use it here (until the PRs get merged upstream).

nix/flake-schemas/flake.nix
srid: > I am not pleased with this approach, let me know if this can be solved better

You could open an issue on upstream repo to initiate discussion among larger audience, as well as propose your `evalOnAllSystems` change as a PR linking to the issue.

flake.nix
shivaraj-bh: PR for flake-schemas input: https://github.com/DeterminateSystems/inspect/pull/6

Discussion about `evalOnAllSystems`: https://github.com/DeterminateSystems/inspect/issues/7

srid:
intel mac builder is stuck for reason, 

image

srid:
> * Credit https://github.com/DeterminateSystems/inspect

I don't know what this entails, but we can do this after this PR is merged, which I'll do now.

crates/nix_rs/src/flake/outputs.rs
```diff
@@ -6,35 +6,71 @@ use std::{
...
+                "--override-input",
+                "flake",
+                flake_url,
+                "--override-input",
+                "systems",
+                // TODO: don't use unwrap
+                &SystemsListFlakeRef::from_known_system(system).unwrap().0,
+                "--no-write-lock-file",
+                // Why `exculdingOutputPaths`?
+                //   This function is much faster than `includingOutputPaths` and also solves <https://github.com/juspay/omnix/discussions/231>
```
srid, line 59: Note to self: I should improve `nix_rs` command invocation API. And make use of this struct,

https://github.com/juspay/omnix/blob/b5f3b54e4a290372889dcceabd8ed48573f05d5f/crates/nix_rs/src/flake/command.rs#L70-L90

srid: #280

crates/omnix-cli/tests/command/show.rs
```diff
@@ -5,7 +5,7 @@ use super::core::om;
 /// `om show` runs, and succeeds for a local flake.
 #[test]
 fn om_show_local() -> anyhow::Result<()> {
-    om()?.arg("show").arg(".").assert().success().stdout(
+    om()?.arg("show").arg("../..").assert().success().stdout(
```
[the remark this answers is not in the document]
shivaraj-bh: `nix flake show` walks up the directory tree to find the flake, while `nix eval` doesn't

PR #286, opened 2024-09-23
nix_rs(nix run): Append `FlakeOptions` after `run` subcommand

shivaraj-bh:
`--override-input` is unrecognised if used before, see:
```
❯ nix --override-input services-flake .. run .#cz-check
error: unrecognised flag '--override-input'
```


srid (approved):
I assume this is the case with the rest of the fields in that struct, like `--no-write-lock-file`.

shivaraj-bh:
Caught in https://github.com/juspay/services-flake/pull/338

image

PR #298, opened 2024-10-07
nix_rs(`nix_eval`): Don't capture stderr

### LINKED ISSUE #209: `om show`: Display progress
`nix flake show` will display download/build progress like:

```
[0/3 built, 7/58/401 copied (137.8/3768.9 MiB), 19.2/406.0 MiB DL] fetching gnutar-1.34 from https://cache.nixoserror: interrupted by the user
```

`om show` on the other hand just goes silent, which can be strange when run on large/complex flakes (with long waiting time).

shivaraj-bh:
Resolves #209 

PRs filed tangential to this:
- https://github.com/NixOS/nix/pull/11652

### DIFF
--- a/crates/nix_rs/src/command.rs
+++ b/crates/nix_rs/src/command.rs
@@ -175,6 +175,8 @@ impl NixCmd {
     }
 
     /// Run Nix with given [Command] customizations, while also tracing the command being run.
+    ///
+    /// Return the stdout bytes returned by [tokio::process::Child::wait_with_output]
     pub async fn run_with<F>(&self, f: F) -> Result<Vec<u8>, CommandError>
     where
         F: FnOnce(&mut Command),
--- a/crates/nix_rs/src/command.rs
+++ b/crates/nix_rs/src/command.rs
@@ -174,46 +174,21 @@ impl NixCmd {
         }
     }
 
-    /// Like [Self::run_with_returning_stdout] but inherits stderr from parent to trace the progress
-    /// of running Nix with [Command] customizations.
-    pub async fn trace_run_with_returning_stdout<F>(&self, f: F) -> Result<Vec<u8>, CommandError>
+    /// Run Nix with given [Command] customizations, while also tracing the command being run.
+    pub async fn run_with<F>(&self, f: F) -> Result<Vec<u8>, CommandError>
     where
         F: FnOnce(&mut Command),
     {
         let mut cmd = self.command();
         f(&mut cmd);
         trace_cmd(&cmd);
-
-        cmd.stdout(Stdio::piped());
-
-        let child = cmd.spawn()?;
-        let out = child.wait_with_output().await?;
-
+        let out = cmd.spawn()?.wait_with_output().await?;
         if out.status.success() {
             Ok(out.stdout)
         } else {
             Err(CommandError::ProcessFailed {
                 stderr: None,
                 exit_code: out.status.code(),
-            })
-        }
-    }
-
-    /// Run Nix with given [Command] customizations, while also tracing the command being run.
-    pub async fn run_with<F>(&self, f: F) -> Result<(), CommandError>
-    where
-        F: FnOnce(&mut Command),
-    {
-        let mut cmd = self.command();
-        f(&mut cmd);
-        trace_cmd(&cmd);
-        let status = cmd.spawn()?.wait().await?;
-        if status.success() {
-            Ok(())
-        } else {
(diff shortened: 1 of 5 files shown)


shivaraj-bh:
image

Tried on `github:nammayatri/nammayatri`

crates/nix_rs/src/command.rs
```diff
@@ -174,6 +174,31 @@ impl NixCmd {
         }
     }
 
+    /// Like [Self::run_with_returning_stdout] but inherits stderr from parent to trace the progress
+    /// of running Nix with [Command] customizations.
+    pub async fn trace_run_with_returning_stdout<F>(&self, f: F) -> Result<Vec<u8>, CommandError>
```
srid, line 179: Can we define this using `run_with` (refactoring the function along the way)?

I wrote `run_with` so that we don't have to create new functions like this.

srid (commented):
Great, 'tis good to see a functional PR ready to refactor/merge.

crates/nix_rs/src/flake/eval.rs
```diff
@@ -13,10 +13,13 @@ where
...
 {
     let stdout = nixcmd
-        .run_with_returning_stdout(|cmd| {
+        .trace_run_with_returning_stdout(|cmd| {
             cmd.args(["eval", "--json"]);
             opts.use_in_command(cmd);
             cmd.arg(url.to_string());
+            // Decrease logging verbosity to avoid polluting the `om` output
+            // with logs other than the flake evaluation progress.
+            cmd.args(["--quiet", "--quiet"]);
```
srid, line 22: Double `--quiet`?

srid: Okay, so it is a repeating option? In that case, `-qq` would be clearer as that's what people normally type on the terminal (like `-vv` for instance).

srid: Also, the second quiet level hides "Added input" output? That's interesting; I wonder if we can do that for other commands (not in this PR).

shivaraj-bh: > Okay, so it is a repeating option? In that case, -qq would be clearer 

unlike `—verbose` there is no shorthand for `—quiet`. From `nix eval —help`:

```sh
      · --quiet Decrease the logging verbosity level.

      · --verbose / -v Increase the logging verbosity level.
```

shivaraj-bh: > Also, the second quiet level hides "Added input" output? That's interesting; I wonder if we can do that for other commands (not in this PR).

Yes, it does

srid: > unlike `--verbose` there is no shorthand for `--quiet`.

Alright then. Sounds like an idea for a PR to nix repo.

crates/nix_rs/src/command.rs
shivaraj-bh: changed `run_with` to remove `trace_run_with_returning_stdout`: https://github.com/juspay/omnix/pull/298/commits/5b83c36ed845d32d5a7d9f897b748cdbd5522d0a

crates/nix_rs/src/flake/eval.rs
shivaraj-bh: https://github.com/NixOS/nix/pull/11652

crates/nix_rs/src/copy.rs
```diff
@@ -27,6 +27,8 @@ pub async fn nix_copy(
     }
     cmd.run_with(|cmd| {
         cmd.args(args);
+        cmd.stdout(Stdio::null());
```
srid, line 30: Is this necessary?

shivaraj-bh: Otherwise stdout will be inherited from the parent process

srid: >  stdout will be inherited from the parent process

Which means the stdout of the process goes to the stdout of the parent process. Isn't this expected behaviour?

shivaraj-bh: Yes, it is. My bad

srid: Compare `main` branch:

https://github.com/juspay/omnix/actions/runs/11222470055/job/31194977412

image

with this PR:

https://github.com/juspay/omnix/actions/runs/11222577781/job/31195313989?pr=298

image

srid (approved):

PR #306, opened 2024-10-08
`om health`: Shell dotfiles are Nix-managed

### LINKED ISSUE #299: `om health`: Check that user's shell is configured in Nix
Add a health check that will ensure that the dotfiles of the user's current shell (normally `zsh` on macOS; `bash` on others) is managed, ultimately, by Nix.

We can do this by:

1. iterating on the list of dotfiles effectively used by `$SHELL`
  - Is there a a way to do this reliably?
1. Checking that those symlinks resolving ultimately to paths under `/nix/store`.

This check should work regardless of home-manager, nix-darwin or NixOS.

rsrohitsingh682:
resolves #299


crates/nix_health/CHANGELOG.md
srid, line 1: Can you revert the irrelevant changes to this file?

crates/nix_health/src/check/shell_configurations.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+fn check_shell_configuration() -> bool {
+    // Define known dotfiles for different shells
+    let mut shell_dotfiles = HashMap::new();
+    shell_dotfiles.insert("zsh", vec![".zshrc"]);
+    shell_dotfiles.insert("bash", vec![".bashrc", ".bash_profile", ".profile"]);
+    // Check if shell configurations (zsh or bash) are managed by Nix
+    are_shell_dotfiles_nix_managed("zsh", &shell_dotfiles)
+        || are_shell_dotfiles_nix_managed("bashrc", &shell_dotfiles)
```
srid, line 37: The logic is wrong here. We want to check the current shell, and current shell only. 

Also, what happens if the user is using a different shell like `fish`?

srid: Also, 'tis time to refactor this ([encapsulate](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming))) by introducing a `struct` along with struct methods.

crates/nix_health/src/check/shell_configurations.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// check if dotfile(s) for a given shell points to /nix/store
+fn are_shell_dotfiles_nix_managed(shell: &str, dotfiles: &HashMap<&str, Vec<&str>>) -> bool {
+    let home = env::var("HOME").unwrap_or_default();
```
srid, line 42: What happens in the *default* case?

crates/nix_health/src/check/shell_configurations.rs
```diff
@@ -0,0 +1,55 @@
...
+            let path = PathBuf::from(&home).join(dotfile);
+            if let Ok(canonical_path) = std::fs::canonicalize(&path) {
+                canonical_path.starts_with("/nix/store/")
+            } else {
+                false
+            }
+        })
+    } else {
+        false
+    }
```
srid, line 54: [Function interface should be small](https://srid.ca/coding#fn-interface) enough to not have dummy code paths like this. I believe this will be addressed by introducing struct.

srid (changes requested):
@rsrohitsingh682 Could you post screenshots of the behaviour on macOS, Linux and on some system where neither bash nor zsh is the current shell?

crates/nix_health/src/check/shell_configurations.rs
rsrohitsingh682: > The logic is wrong here. We want to check the current shell, and current shell only.

@srid I tried using `SHELL` environment variable  for this but it turns out it gives the default shell of the user not the current active shell. There are other ways to determine the current active shell but those commands doesn't work well with rust. Hence I went ahead with this method for now.

Screenshot 2024-10-09 at 7 04 15 PM

srid: @rsrohitsingh682 Do you understand why the logic is wrong? If I have both bash and zsh dotfiles, and bash dotfiles are managed by Nix but zsh dotfiles are not managed zsh - then what would be the result of the check if zsh is my current shell?

crates/nix_health/src/check/shell_configurations.rs
rsrohitsingh682: It wouldn't panic instead will return empty string. I didn't want to throw error here because it wouldn't let the health process to be completed (break in between). Hence I thought if it returns empty string we would eventually show that shell configuration is not managed by nix ( However this is still not correct way).Would need your advice here.

srid: [We should not reflexively use Default instance without understanding what it does.](https://github.com/juspay/omnix/pull/224).

Do you understand how a) an empty string for $HOME is meaningless, and b) how $HOME being empty affects the logic in your function? When (b) happens, is your function still correct?

crates/nix_health/src/check/shell_configurations.rs
rsrohitsingh682: oh! Right!! Got it. I made the check to use `OR` condition thinking what if user is just using only one shell either of (zsh or bash). But yeah I should have considered the case that you mentioned.

rsrohitsingh682:
> @rsrohitsingh682 Could you post screenshots of the behaviour on macOS, Linux and on some system where neither bash nor zsh is the current shell?

MacOS where zsh is managed by nix.

Screenshot 2024-10-10 at 2 35 34 PM


MacOS where zsh is not managed by nix.

macOS


Linux (Ubuntu) where bash is not managed by nix.

Ubuntu

crates/nix_health/src/check/shell_configurations.rs
```diff
@@ -0,0 +1,97 @@
...
+    #[serde(other)]
+    Unknown,
+}
+
+impl Shell {
+    fn from_path(path: &str) -> Self {
+        let shell_name = Path::new(path)
+            .file_name()
+            .and_then(|name| name.to_str())
+            .unwrap_or("");
```
srid, line 22: Why do you think empty string is better than `None` (in the context of the `match` below)? What do you think is the purpose of the `Option` type?

srid (commented):
I think the struct encapsulation can still be improved (for one thing, you only need one struct not two). But the best way to understand why is to answer this question: How does this behave under the following situations?

- User's shell is neither bash nor zsh. Its dotfiles are not managed by Nix.
- User's shell is neither bash nor zsh. Its dotfiles are managed by Nix.

crates/nix_health/src/check/shell_configurations.rs
rsrohitsingh682: That's a miss. I didn't realise I could write it like this as well.
```
fn from_path(path: &str) -> Self {
        Path::new(path)
            .file_name()
            .and_then(|name| name.to_str())
            .map(|shell_name| match shell_name {
                "zsh" => Shell::Zsh,
                "bash" => Shell::Bash,
                _ => Shell::Unknown,
            })
            .unwrap_or(Shell::Unknown)
    }

**Comment by rsrohitsingh682** [comment]
> I think the struct encapsulation can still be improved (for one thing, you only need one struct not two). But the best way to understand why is to answer this question: How does this behave under the following situations?
> 
> * User's shell is neither bash nor zsh. Its dotfiles are not managed by Nix.
> * User's shell is neither bash nor zsh. Its dotfiles are managed by Nix.

For both the cases it would return `false` and say it is not managed by Nix which would be wrong for the second part. Now when I think about it I am checking only dotfiles but for shells like fish the configurations are present in `~/.config/fish/config.fish`. So ideally I should maintain a list where I can refer to the correct place be it .dotfiles or config files.

File: crates/nix_health/src/check/shell_configurations.rs (hunk 4 above)
Comment (srid) [reply crates/nix_health/src/check/shell_configurations.rs:22]: It still is not faithful to the environment. Are you really sure `Shell::Unknown` can never refer to when the user has bash or zsh as their current shell? In your code snippet above, there is in fact a condition under which your function will return `Shell::Unknown` despite the user having bash or zsh as their current shell.

Try to understand what each function does, including the nature of the range of values it can return.

Comment (srid) [reply crates/nix_health/src/check/shell_configurations.rs:22]: Overall: we need a case to handle unexpected situations and errors, and then report that accordingly instead of giving a false negative.

**Comment by rsrohitsingh682** [comment]
@srid The CI seems to be failing for `aarch64-darwin` while running tests.

For checking the shell configuration it is looking into 

`/var/lib/github-runners/sambar-juspay-08/.bashrc`

Any thoughts on this ?

Screenshot 2024-10-24 at 3 01 47 PM

**Comment by srid** [comment]
That's because in CI `$HOME` is `/var/lib/github-runners/sambar-juspay-08/`.

How are you handling the case of a dotfile not even existing in the first place?

**Comment by rsrohitsingh682** [comment]
> That's because in CI `$HOME` is `/var/lib/github-runners/sambar-juspay-08/`.
> 
> How are you handling the case of a dotfile not even existing in the first place?

https://github.com/juspay/omnix/pull/306/files#diff-f499d0e179c61c3e80fd0d383c0d1574dd411dba6cb707c91bde716d75308709R129

**Comment by srid** [comment]
Should a missing dotfile cause omnix to panic, or should it cause health check to fail?

**Comment by rsrohitsingh682** [comment]
> Should a missing dotfile cause omnix to panic, or should it cause health check to fail?

If I am able to determine user's default shell, in this case I am and after that if I am not able to find dotfile (in this case because of permission issue I guess), omnix should panic.

From our last discussion, 
if user's shell is not supported then I can cause the health check to fail and for other reasons like $HOME or $SHELL is not set then it should panic. And I think not able to find dotfile is could also be  the reason to panic. I mean this is my understanding, I could be wrong though.

**Comment by rsrohitsingh682** [comment]
> A missing dotfile is a legitimate use case. It should fail the health check, rather than panic. Then, if this check is required=false, the CI would not fail.

Got it. I will make those changes then.

File: crates/omnix-health/src/check/shell.rs [hunk 5]
```diff
@@ -0,0 +1,158 @@
...
+use crate::traits::{Check, CheckResult, Checkable};
+
+/// Shell types
+///
+#[derive(Debug, Serialize, Deserialize, Clone, Hash, Eq, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Shell {
+    Zsh,
+    Bash,
+    Undeterminable,
```
srid, line 13: What does `Undeterminable` actually mean? I don't think should be treated as a shell; it should be an error!

crates/omnix-health/src/check/shell.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[serde(other)]
+    Other,
```
srid, line 15:
```suggestion
    /// An unknown string located at given path
    Other(String),
```

crates/omnix-health/src/check/shell.rs
```diff
@@ -0,0 +1,158 @@
...
+            Shell::Bash => "bash",
+            Shell::Other => "<unsupported shell>",
+            Shell::Undeterminable => "<undeterminable shell>",
+        };
+        write!(f, "{}", shell_str)
+    }
+}
+
+impl Shell {
+    /// Creates a `Shell` from a path string.
```
srid, line 31: This documentation is implement. What kind of paths are this function expected to take? Can I pass, for example, `"/tmp"`?

crates/omnix-health/src/check/shell.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn from_path(path: &str) -> Self {
```
srid, line 32: Why does the function say *path*, but takes a *string* (instead of `Path`) instead?

crates/omnix-health/src/check/shell.rs
```diff
@@ -0,0 +1,158 @@
...
+        let shell_name = Path::new(path).file_name().and_then(|name| name.to_str());
```
srid, line 33: https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name

None is a candidate for runtime error.

crates/omnix-health/src/check/shell.rs
```diff
@@ -0,0 +1,158 @@
...
+            ]),
+            Shell::Other => Err(ShellError::UnsupportedShell),
+            Shell::Undeterminable => Err(ShellError::UndeterminableShell),
+        }
+    }
+}
+
+impl Default for Shell {
+    fn default() -> Self {
+        let shell_path = std::env::var("SHELL").expect("Environment variable `SHELL` not set");
```
srid, line 60: make it a `PathBuf` directly here.

crates/omnix-health/src/check/shell.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// Checks if all dotfiles for a given shell are managed by nix
+///
+/// # Returns
+/// * `true` if all dotfiles are nix-managed
+/// * `false` if any dotfile is not nix-managed
+/// * `Err` if there was an error during the check
+fn are_dotfiles_nix_managed(shell: &Shell) -> Result<bool, ShellError> {
+    let home_dir = PathBuf::from(
+        std::env::var("HOME").map_err(|_| ShellError::EnvVarNotSet("HOME".to_string()))?,
```
srid, line 115: No need to handle this. `expect` just like for $SHELL

srid:
@rsrohitsingh682 left my last sequence of reviews; after this, I'll try to evaluate/refactor before merging.

srid:
> @rsrohitsingh682 Could you post screenshots of the behaviour on macOS, Linux and on some system where neither bash nor zsh is the current shell?

Answer: behaviour when using `fish`:

image

rsrohitsingh682:
> > @rsrohitsingh682 Could you post screenshots of the behaviour on macOS, Linux and on some system where **neither bash nor zsh** is the current shell?
> 
> Answer: behaviour when using `fish`:
> 
> image

This was because of the below commit.

https://github.com/juspay/omnix/pull/306/commits/5aaa25ef78ebe5a3242e5a6e6d0f8bbfc7c340b7#diff-f499d0e179c61c3e80fd0d383c0d1574dd411dba6cb707c91bde716d75308709R95-R97

srid:
@rsrohitsingh682 Yes, this is how we want it to behave.

srid (approved):

Issue #304, opened 2024-10-07
CI rate limit issue


srid:
<img width="710" alt="image" src="https://github.com/user-attachments/assets/f6cb68fc-8067-4bc4-83a5-b374e612b847">

---

```
❯ curl -q -I https://api.github.com/repos/srid/haskell-flake/commits/master | grep x-ratelimit
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0   278    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
x-ratelimit-limit: 60
x-ratelimit-remaining: 0
x-ratelimit-reset: 1728335386
x-ratelimit-resource: core
x-ratelimit-used: 60
```

https://github.com/NixOS/nix/issues/5409

PR #308, opened 2024-10-09
Fix build on `aarch64-linux`; Build for `aarch64-linux` in CI

### LINKED ISSUE #287: Omnix fails to compile on `aarch64-linux`
image

shivaraj-bh:
resolves #287

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,7 +25,7 @@ jobs:
       OMWEBSITE: ${{ steps.omci.outputs.OMWEBSITE_x86_64-linux }}
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


srid (approved):
Great; we can merge this after https://github.com/juspay/nixos-buffet/pull/40 is deployed and CI is re-triggered.

srid:
CI failure now due to #304 - will have to try again in an hour or so.

srid (changes requested):
The build's failing. Note: you can run `om ci` (or `just ci-cargo`) to reproduce CI locally before pushing.

image

srid:
We could also disable builds of `nixci` ad `nix_health` on aarch64-linux and x86_64-darwin to save on CI time (by null'ifying their `autoWire`). We can do doc and clippy checks on primary architectures only.

PR #312, opened 2024-10-09
init: Make registry pure

srid:
The registry flakes are now locked in a flake.lock file.

Because we use flake=false, the lockfile is now manageable.

This should substantially reduce #304

### DIFF
--- a/crates/omnix-init/registry/flake.nix
+++ b/crates/omnix-init/registry/flake.nix
@@ -1,0 +1,26 @@
+# `om init` registry's canonical source
+#
+# Get JSON using:
+# nix eval --json .#registry | jq
+{
+  inputs = {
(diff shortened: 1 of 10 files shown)


crates/omnix-init/registry/flake.nix
```diff
@@ -0,0 +1,26 @@
...
+
+    nixos-unified-template.url = "github:juspay/nixos-unified-template";
+    nixos-unified-template.flake = false;
+  };
+
+  outputs = inputs: {
+    registry =
+      builtins.mapAttrs
+        (k: v: v.outPath)
+        (builtins.removeAttrs inputs [ "self" ]);
```
srid, line 24: We can potentially use flake-schemas here in the future. Not sure what value it will add, though.

PR #324, opened 2024-10-18
Use `--quiet --quiet` by default for `om ci` and `om show`

### LINKED ISSUE #300: Use `--quiet --quiet` wherever possible to Nix
>> Also, the second quiet level hides "Added input" output? That's interesting; I wonder if we can do that for other commands (not in this PR).

> Yes, it does

_Originally posted by @shivaraj-bh in https://github.com/juspay/omnix/pull/298#discussion_r1790622178_
            

image

shivaraj-bh:
resolves #300


shivaraj-bh:
I just realised, `om ci` might not need it yet, because `devour_flake.rs` pipes both `stdout` and `stderr`
https://github.com/juspay/omnix/blob/be6ce4a0c31fc4cde860104baeb7af02ae924c73/crates/nixci/src/nix/devour_flake.rs#L77

srid:
In situations do we need this done for `om show`?

shivaraj-bh:
> In situations do we need this done for `om show`?

When `nix_eval` is called trying to fetch flake schemas:
https://github.com/juspay/omnix/blob/be6ce4a0c31fc4cde860104baeb7af02ae924c73/crates/nix_rs/src/flake/schema.rs#L84

shivaraj-bh:
On a second thought, it would be more appropriate to localise modifying `NixCmd` with `--quiet --quiet` right before where it is needed.

srid:
> > In [what] situations do we need this done for `om show`?
> 
> When `nix_eval` is called trying to fetch flake schemas:

Not sure I understand. I don't see any override input relate logs when running, for example, 
image

shivaraj-bh:
The logs from `--override-input` are not shown in `om show` because we add `--quiet --quiet` by default in [nix_eval](https://github.com/juspay/omnix/blob/769d7b78998c7935ef34677d35143e88ef382427/crates/nix_rs/src/flake/eval.rs#L23-L24). This PR allows to append `--quiet --quiet` directly to the `NixCmd`, instead of being specific to `nix_eval` and thus expanding the scope to use it in `NixCmd` used during `om ci`.

crates/nix_rs/src/command.rs
```diff
@@ -119,6 +124,13 @@ impl NixCmd {
         cmd
     }
 
+    /// Suppress logs related to `override-input` usage by reducing `nix` command's verbosity
```
srid, line 127: Is this (_Suppress logs related to `override-input` usage_) the **only** effect of using `--quiet --quiet` or does it suppress anything else?

srid: If it does more than that, then function (`mute_override_input_logs`) wouldn't be accurate.

shivaraj-bh: Logically, it reduces verbosity level of the logs. I could rename the function to depict that.

crates/nix_rs/src/command.rs
```diff
@@ -40,6 +40,10 @@ pub struct NixCmd {
     #[cfg_attr(feature = "clap", arg(long))]
     pub extra_access_tokens: Vec<String>,
 
+    /// Arguments to pass verbatim to the Nix command
+    #[cfg_attr(feature = "clap", arg(last = true))]
+    pub extra_args: Vec<String>,
```
srid, line 45: Instead of `extra_args`, just add a `quiet` field looking like this. It is more self-documenting.

https://github.com/clap-rs/clap-verbosity-flag/blob/a7e305e8b0bde21be222ae658cc14813464ea3e6/src/lib.rs#L79-L88

srid (changes requested):
@shivaraj-bh I tried out this PR locally and there appears to be a bug.

Basically it suppresses the entire output of `nix build` (I see no build logs):

Compare this branch:

image

vs main:

image

srid:
Closing per https://github.com/juspay/omnix/issues/300#issuecomment-2465532160

PR #341, opened 2024-11-08
Read configuration once only

shivaraj-bh:
Before
```sh
⌨️  Preparing to develop project: .
❄️  nix --version️
❄️  nix --extra-experimental-features nix-command show-config --json️
❄️  nix eval '.#om.develop' --json️
❄️  nix eval '.#om.health' --json️
✅ Nix environment is healthy.

🍾 Welcome to the omnix project
```
After
```sh
❯ just run develop
❄️  nix --version️
❄️  nix --extra-experimental-features nix-command show-config --json️
❄️  nix eval '.#om' --json️
⌨️  Preparing to develop project: .
✅ Nix environment is healthy.

🍾 Welcome to the omnix project
```

Notice that we only evaluate the `om` config once in: `nix eval '.#om' --json️`


srid:
`omnix-common` now exports two (instead of one) public functions. Why do we need two rather one? If we do need two, when should we use one over the other? What is the difrerence between them? I also ask, because the code comment does not clarify this for me.

image

image

shivaraj-bh:
> `omnix-common` now exports two (instead of one) public functions. Why do we need two rather one? If we do need two, when should we use one over the other? What is the difrerence between them? I also ask, because the code comment does not clarify this for me.
> 
> 
> 
> image
> 
> 
> 
> image
> 
> 

get_sub_configs is for cases like om.templates. It is also used in om.develop, but I guess we could use get_referenced_for there. And get_referenced_for is when it makes for user to pick one amongst many possible sub-configs.

srid:
Do I understand correctly one is same as other, except for returning multiple (instead of one) keys?

In that case, shouldn't `get_referenced_for` be defined in terms of (using) `get_sub_configs` since the latter is a superset of the former?

shivaraj-bh:
Yes, but there is also that get_referenced_for requires the type T, that it is converting to, have a Default defined, but that is not the case in get_sub_config.

srid:
I mean, why can't `get_referenced_for` be simply defined as (roughly)  `get_sub_configs().get("default").unwrap_or_default()` instead of the 28 line function body that it is?

srid:
Basically, I think `pub config: BTreeMap<String, BTreeMap<String, serde_json::Value>>,` should newtype'd as separate struct, with its own impl methods are operate just on the BTreeMap without regard to other fields (like `reference`). 

Decoupled, basically.

srid (changes requested):
I've tested this locally and it works. And I'm glad that this simplifies the code by consolidating config in one place, and getting rid of qualified_attr.rs (it really was complicated solution to begin with!).

Could you do the decoupling mentioned in comments? Then it can be merged.

shivaraj-bh:
> I've tested this locally and it works. And I'm glad that this simplifies the code by consolidating config in one place, and getting rid of qualified_attr.rs (it really was complicated solution to begin with!).
> 
> Could you do the decoupling mentioned in comments? Then it can be merged.

I did the decoupling, although not super happy with the name `Config` for `pub config: BTreeMap<String, BTreeMap<String, serde_json::Value>>` , but couldn’t think of anything better.

srid (approved):

PR #345, opened 2024-11-15
`om develop`: Use `FlakeUrl` without attrs for fetching `OmConfig`

shivaraj-bh:
As was the case before https://github.com/juspay/omnix/pull/341

### DIFF
--- a/crates/omnix-cli/src/command/develop.rs
+++ b/crates/omnix-cli/src/command/develop.rs
@@ -28,7 +28,7 @@ impl DevelopCommand {
     pub async fn run(&self) -> anyhow::Result<()> {
         let flake = self.flake_shell.without_attr();
 
-        let om_config = OmConfig::from_flake_url(NixCmd::get().await, &self.flake_shell).await?;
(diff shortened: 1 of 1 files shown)


crates/omnix-cli/src/command/develop.rs
```diff
@@ -28,7 +28,7 @@ impl DevelopCommand {
     pub async fn run(&self) -> anyhow::Result<()> {
         let flake = self.flake_shell.without_attr();
 
-        let om_config = OmConfig::from_flake_url(NixCmd::get().await, &self.flake_shell).await?;
+        let om_config = OmConfig::from_flake_url(NixCmd::get().await, &flake).await?;
```
srid, line 31: I'm confused. This change impacts behaviour? `from_flake_url` already strips attr anyway. See 

image

and:

image

srid: Oh I get it, the attr is expected to be config path, which is not always the case (as with nammayatri `.#backend`).

PR #356, opened 2024-12-02
Fix: `just cargo-test` was erroneously running `cargo doc`

shivaraj-bh:
Also fixes tests failing due to deprecation of `nixci` and `nix-health` flake configs. See
https://github.com/juspay/omnix/blob/df0093881be956a9df8034e5970dc363838bf478/doc/src/history.md?plain=1#L33

Notes on changes:
- `test_services_flake` doesn’t run `om ci` on all the sub-flakes (which was the case before) because unlike before we have a lot of services and since the goal of the test is to just check if the `stdout` is not polluted with non-outPaths, testing a smaller sub-flake should suffice.
- `test_config_loading` upgrades the commit of `haskell-flake` to the one that uses `om.ci` configuration.

### DIFF
--- a/crates/omnix-ci/src/config/core.rs
+++ b/crates/omnix-ci/src/config/core.rs
@@ -10,15 +10,15 @@ mod tests {
     #[tokio::test]
     async fn test_config_loading() {
         // Testing this flake:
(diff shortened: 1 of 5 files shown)


shivaraj-bh:
https://github.com/juspay/omnix/actions/runs/12117052416/job/33778697244?pr=356
Confirmed that the tests are now running in CI

image

Issue #358, opened 2024-12-09
`om ci`: `-o` does not work when used with `--on`


srid:
Repro:

```
om ci run --on ssh://pureintent -o om.json
```

This will build the flake remotely on `pureintent`, however the `-o` option is ignored.

### Ideal outcome

Have `-o om.json` produce that file *locally*.

srid:
Somehow related: I think there should be a way to have `om ci run --on` copy back all the built store paths to a local nix store. Otherwise, the paths in this `om.json` would be meaningless. Overall, this would also obviate remote builders - where the boundary is not very clear.

PR #359, opened 2024-12-09
`om ci`: Add devour-flake output path in result JSON

srid:
Useful to `nix copy` stuff back transitively (https://github.com/juspay/omnix/issues/358#issuecomment-2529325175).

image


srid:
image

PR #361, opened 2024-12-11
Pass `--accept-flake-config` only during flake-related commands

shivaraj-bh:
`--accept-flake-config` expects the flake to be already enabled, which will fail `om health` on systems with flake not enabled.

## Error

```
❄️  nix --accept-flake-config --version️
Error: Unable to gather nix info

Caused by:
    0: Nix command error: Command error: Process exited unsuccessfully. exit_code=Some(1) stderr=error: experimental Nix feature 'flakes' is disabled; add '--extra-experimental-features flakes' to enable it

    1: Command error: Process exited unsuccessfully. exit_code=Some(1) stderr=error: experimental Nix feature 'flakes' is disabled; add '--extra-experimental-features flakes' to enable it

    2: Process exited unsuccessfully. exit_code=Some(1) stderr=error: experimental Nix feature 'flakes' is disabled; add '--extra-experimental-features flakes' to enable it
```

### DIFF
--- a/crates/nix_rs/CHANGELOG.md
+++ b/crates/nix_rs/CHANGELOG.md
@@ -54,7 +54,7 @@
 - **`eval`**
   - `nix_eval_attr_json`
     - No longer takes `default_if_missing`; instead (always) returns `None` if attribute is missing.
-    - Rename to `nix_eval_attr` (as there is no non-JSON variant)
+    - Rename to `nix_eval_maybe` (as there is no non-JSON variant)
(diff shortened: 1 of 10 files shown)


shivaraj-bh:
> I think we should just make the default `accept_flake_config` to be false, and instead set it to true in `with_flakes()`.

Done

srid:
I just realized there's another bug in `main` ... `--accept-flake-config` is not even being passed in many places! Fixing this as part of this PR ...

Issue #377, opened 2025-01-14
`om ci run —on` fails if flake input is private repo


rsrohitsingh682:
When using `om ci run —on` option to run ci on a remote machine, the expectation is all flake inputs should be copied to the remote machine.

But it fails when flake input is of private repo.

<img width="1724" alt="Screenshot 2025-01-14 at 8 37 03 PM" src="https://github.com/user-attachments/assets/0a8ae1f4-ae93-4f38-9c07-c83e2f83713e" />

Steps to reproduce:

`nix run github:juspay/omnix — ci run --on ssh://<user>@<ip> github:rsrohitsingh682/haskell-repo`

srid:
Ah, omnix never copied the private repo:

<img width="803" alt="Screenshot 2025-01-14 at 10 33 24 AM" src="https://github.com/user-attachments/assets/3401d815-caec-4ca9-b1ed-df876b387ac1" />

So `nix copy`'ing a flake does not automatically copy over its flake inputs.

srid:
The solution is to move away from `nix flake metadata` to something else (an expression defined in Nix, just like `devour-flake`) for retrieving flake metadata including outPath's of flake inputs.

srid:
Resolved in #381 

Use `--on ssh://host?copy-inputs=true`

https://omnix.page/om/ci.html#remote

PR #380, opened 2025-01-16
Generalize `devour-flake`'s FFI mechanism

srid:
This prepares the ground for https://github.com/juspay/omnix/issues/377#issuecomment-2596911041


srid:
Also a more generalized pattern than https://github.com/DeterminateSystems/inspect

PR #387, opened 2025-01-20
nix_rs: Informative "Unknown installer"

### LINKED ISSUE #382: `om health`: Make "Unknown installer" more informative
Perhaps add the full output of `which nix` to the message. I wonder, if in this screenshot, the user had installed Nix through `apt-get`.

Image

shivaraj-bh:
resolves #382 

Tested with official Nix installer on macOS VM:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1673,6 +1673,7 @@ dependencies = [
  "tokio",
  "tracing",
  "url",
+ "which",
  "whoami",
 ]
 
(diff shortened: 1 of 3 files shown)


srid:
This is useful change, but I was hoping to also deal with https://github.com/juspay/omnix/issues/382#issuecomment-2598540647=

Sourish and his coworker was using Ubuntu here. I'm guessing the new nix-installer is no longer creating `/nix/nix-installer`, at least on certain platforms? If so, we should change our DetSys installer detection to account for it.

shivaraj-bh:
>  I'm guessing the new nix-installer is no longer creating /nix/nix-installer, at least on certain platforms?

Wasn’t able to reproduce on Ubuntu or macOS VMs, hence decided to go ahead without considering this edge case.

PR #398, opened 2025-01-22
`om health`: Add `supported` nix versions health check

### LINKED ISSUE #386: `om health`(`nix-version`): Add `vulnerableVersions`
For times when only specific versions of `nix` are problematic

shivaraj-bh:
resolves #386 

## Changes
- Introduces version specification system for `NixVersion`
- Replaces `nix-versions.min-required` health check with more flexible `nix-versions.supported`

The new version specification system provides APIs to parse complex version requirements, enabling fine-grained control over which Nix versions should pass the health check.

Example of a version requirement string for `nix-versions.supported`:
```
“>=2.16.0, !=2.23.10, <2.25"
```
This requirement specifies that:
- Versions must be at least 2.16.0
- Must not be exactly 2.23.10 
- Must be less than 2.25

Any Nix version meeting all these criteria will pass the health check.
  
## TODO
- [x] Add tests
- [x] Do we still have to support `nix-version.min-required`? Even if we are to deprecate it, how to do it?
- [x] https://github.com/juspay/omnix/pull/398#discussion_r1928524329
- [x] ~~Re-check the implementation of `version_spec::NixVersionReq::check` as it doesn’t yet account for `Gteq` and `Lteq`~~ 
- [x] A blog post on various ways to get a specific version of `nix` and link it as a `Fix` when `Supported Nix versions` fails? (https://github.com/nixos-asia/website/pull/47)


srid:
> * Do we still have to support `nix-version.min-required`? Even if we are to deprecate it, how to do it?

Depends on the behaviour when we run this against old code. What is the output in that case?

shivaraj-bh:
>Depends on the behaviour when we run this against old code (which specifies min-required). What is the output in that case?

It is silently ignored.

shivaraj-bh:
An example of failure using the new NixVersion spec:

image

crates/omnix-health/src/check/nix_version.rs
```diff
@@ -0,0 +1,54 @@
...
+use nix_rs::info;
+use serde::{Deserialize, Serialize};
+
+use crate::traits::*;
+
+/// Check that [nix_rs::version::NixVersion] is set to a good value.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+#[serde(default, rename_all = "kebab-case")]
+pub struct NixVersionCheck {
+    pub supported: String,
```
shivaraj-bh, line 12: I wish serde supported passing a function as a deserializer. I could just use `pub supported: NixVersionReq` here, if it did. Maybe it does and I haven’t looked properly?

shivaraj-bh: Maybe `#[serde(deserialize_with = “<function-name>”)]`?

shivaraj-bh: https://github.com/juspay/omnix/pull/398/commits/53355baad21ed7cdaaa0e21f171d8ffb7215f800

works

crates/nix_rs/src/version.rs
```diff
@@ -10,14 +10,14 @@ use tracing::instrument;
...
+#[derive(Clone, Copy, Debug, SerializeDisplay, DeserializeFromStr)]
 pub struct NixVersion {
     /// Major version
     pub major: u32,
     /// Minor version
-    pub minor: u32,
+    pub minor: Option<u32>,
     /// Patch version
-    pub patch: u32,
+    pub patch: Option<u32>,
```
srid, line 20: This changes the semantics of this type. If you look at the type documenation, it says 

```
Nix version as parsed from `nix --version`
```

And `nix --version` always returns all 3 components - major, minor and patch.

---

It seems that the type here is being changed to fit something in `version_spec.hs` - yet `version.rs` is isolated from that module.

crates/nix_rs/src/version_spec.rs
```diff
@@ -0,0 +1,204 @@
+//! Rust module defining the spec for [NixVersion]
```
srid, line 1:
```suggestion
//! Version requirement spec for [NixVersion]
```

crates/nix_rs/src/version_spec.rs
```diff
@@ -0,0 +1,204 @@
...
+    Gteq(NixVersion),
+    /// Version must be less than the specified version
+    Lt(NixVersion),
+    /// Version must be less than or equal to the specified version
+    Lteq(NixVersion),
+    /// Version must not equal the specified version
+    Neq(NixVersion),
+}
+
+/// A collection of version requirements
```
srid, line 25:
```suggestion
/// Version requirement for [NixVersion]
///
/// Example:
/// ```
/// >=2.8, <2.14, 12.13.4
/// ```
```

crates/nix_rs/src/version_spec.rs
```diff
@@ -0,0 +1,204 @@
+//! Rust module defining the spec for [NixVersion]
+
+use std::{fmt, str::FromStr};
+
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::version::{BadNixVersion, NixVersion};
+
+/// Specifies a version requirement for Nix
```
srid, line 10:
```suggestion
/// An individual component of `NixVersionReq`
```

crates/nix_rs/src/version_spec.rs
```diff
@@ -0,0 +1,204 @@
...
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Self::check(&specs)?;
+        Ok(NixVersionReq { specs })
+    }
+
+    /// Checks if all version specifications are satisfiable
+    ///
+    /// Returns Err if the requirements are logically impossible to satisfy
+    fn check(specs: &[NixVersionSpec]) -> Result<(), BadNixVersionSpec> {
```
srid, line 70: I was confused by this function for a moment.

```suggestion
    fn ensure_logically_sound(specs: &[NixVersionSpec]) -> Result<(), BadNixVersionSpec> {
```

crates/omnix-health/src/check/nix_version.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Check that [nix_rs::version::NixVersion] is set to a good value.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+#[serde(default, rename_all = "kebab-case")]
+pub struct NixVersionCheck {
+    #[serde(deserialize_with = "deserialize_version_req")]
+    pub supported: NixVersionReq,
+}
+
+fn deserialize_version_req<'de, D>(deserializer: D) -> Result<NixVersionReq, D::Error>
```
srid, line 16: This?

https://docs.rs/serde_with/latest/serde_with/derive.DeserializeFromStr.html

crates/nix_rs/src/version.rs
shivaraj-bh: Reverted changes to `version.rs`, the edge case of not having `minor`, `patch` or both `minor` and `patch` components, is now handled by `version_spec.rs`

crates/nix_rs/src/version_spec.rs
shivaraj-bh: I have now removed this function, as per our huddle discussion, to keep the interface simple.

crates/nix_rs/src/version_spec.rs
```diff
@@ -0,0 +1,204 @@
...
+impl fmt::Display for NixVersionReq {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut specs = self.specs.iter();
+
+        if let Some(first) = specs.next() {
+            write!(f, "{}", first)?;
+            for spec in specs {
+                write!(f, ", {}", spec)?;
+            }
+        }
```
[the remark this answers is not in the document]
shivaraj-bh: Done

crates/omnix-health/src/check/nix_version.rs
shivaraj-bh: Switched to `DeserializeFromStr`

PR #399, opened 2025-01-22
Track dependencies in`om ci`'s result symlink

### LINKED ISSUE #390: Use `builtins.storePath` to make omnix `result` file with dependencies?
https://nix.dev/manual/nix/2.23/language/string-context

Image

srid:
Resolves #390

- `om ci run`'s `result` symlink now behaves like `nix build`'s result symlink in that all the `outPaths` in it are tracked as dependency. 
  - Since we use nix_rs flake "functions" (which in turn uses `nix build`), these symlinks are automatically pinned in gcroots, `/nix/var/nix/gcroots/auto`
  - `--on` now copies just this result, without needing to explicitly copy the paths
  - Remove `?copy-output=`, because that is now implicitly handled by `om ci run --no-out-link`. Results are copied if out-link is requested (default), or not.


crates/nix_rs/src/flake/functions/core.rs
```diff
@@ -50,6 +50,10 @@ pub trait FlakeFn {
     fn call(
         nixcmd: &NixCmd,
         verbose: bool,
+        // FIXME: Don't do this; instead take dyn trait options
```
srid, line 53: Have to address this in separate PR.

srid:
I've deployed https://github.com/juspay/omnix/commit/72ef90cfbe136270ed37c98a66d1d5fc7249301d to OSS runners.

@shivaraj-bh Feel free to deploy to private runners as/when needed. Now `om ci run` will generate a `result` link with entire closure of built paths, which means - they won't be GC'ed (due to pinning in `/nix/var/nix/gcroots/auto`) as long as the CI working copy remains, which is the case with Jenkins (if not github runners).

PR #413, opened 2025-02-07
feat(ci): improve grouped logging output

### LINKED ISSUE #234: `om ci`: Group CI output for github actions
## Problem

After #232, our CI output is pretty monolithic, and hard to distinctly make out each steps. Notice, in [this](https://github.com/juspay/omnix/actions/runs/13164151812/job/36740068437) example, how all output of `om ci run` is grouped under this `nix run` step:

Image

## What we want

We want each `om ci` step defined here:

https://github.com/juspay/omnix/blob/79e4386882be3f41fd97b5c68c456a67f1a82355/om.yaml#L5

to be [grouped](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines) in Github Actions log output.

## Approach

Consider implementing a `om ci run --github-output` option or some such thing that will enable grouping in GitHub UI.

See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines

niksingh710:
Refactored grouped logging to avoid redundant checks for GitHub Actions output.
Ensured cleaner and more structured execution for CI steps.

Fixes #234


srid (commented):
This seems to work only partially? I see some log lines being sent outside of their groups:

image

https://github.com/juspay/omnix/actions/runs/13202017280/job/36856054936?pr=413

srid:
> fix(ci)

FYI: This is a *feature* not a bug fix.

srid (commented):
This looks good. The only thing I'd add is to make `--github-output` true by default if it is run in github CI (such as, when `GITHUB_ACTION` env var is set).

srid (commented):
The tests are failing now. This is probably because you are printing to stdout, which is expected to contain only the store paths. Using `stderr` should be sufficient, I think - but I'm unsure if GitHub Actions will pick it up as grouping marker.

image

niksingh710:
Performed a push sending the group string output to `stderr`, but at https://github.com/juspay/omnix/pull/413/commits/0fd23a831eb379007554d1b57264841c2458b8f2 workflow executed fine.

it breaks after https://github.com/juspay/omnix/pull/413/commits/a6706fad85cb7a2c99caad226ea64f18e47b71cc the introduction of `GITHUB_ACTION` handler.

crates/omnix-ci/src/command/run.rs
```diff
@@ -220,6 +233,7 @@ pub async fn ci_run(
     run_cmd: &RunCommand,
     cfg: &OmConfig,
     nix_config: &NixConfig,
+    github_output: bool,
```
srid, line 236: You don't need this argument, because that information is already in the `run_cmd` argument.

crates/omnix-ci/src/command/run.rs
```diff
@@ -249,12 +263,20 @@ pub async fn ci_run(
...
         tracing::info!("\n🍎 {}", name);
         let steps_res = subflake
             .steps
             .run(cmd, verbose, run_cmd, &systems, &cfg.flake_url, subflake)
             .await?;
         res.insert(subflake_name.clone(), steps_res);
+
+        if github_output {
+            eprintln!("::endgroup::");
+        }
```
srid, line 279: You should consider a `with_*` style function, as in:

```rust
in_github_group(format!("Running {}", name), move async || {
  subflake.steps.run(...
});

srid (commented):
I'd like to merge, but just a few more things to do.

Also, could you use the `in_github_group` functions to group the whole health check part here?

image

srid (approved):
Looks great,

image

PR #415, opened 2025-02-10
feat(schema): Added schema for process-compose. [debug-based]

niksingh710:
* Fully depended on debug flag of flake-parts.

To test make sure that the `debug` flag is true in a flake-parts flake.

e.g on [`juspay/services-flakes/example/llm`](https://github.com/juspay/services-flake/tree/main/example/llm) apply patch
```patch
diff --git a/example/llm/flake.nix b/example/llm/flake.nix
index c7ce558..a5cc377 100644
--- a/example/llm/flake.nix
+++ b/example/llm/flake.nix
@@ -8,6 +8,7 @@
     services-flake.url = "github:juspay/services-flake";
   };
   outputs = inputs: inputs.flake-parts.lib.mkFlake { inherit inputs; } {
+    debug = true;
     systems = import inputs.systems;
     imports = [
       inputs.process-compose-flake.flakeModule
```

### DIFF
--- a/crates/omnix-cli/src/command/show.rs
+++ b/crates/omnix-cli/src/command/show.rs
@@ -146,6 +146,14 @@ impl ShowCommand {
                 self.flake_url
             )),
         );
+        print_flake_output_table(
+            "🔄 Process Compose",
+            &["allSystems", system.as_ref()],
+            Some(format!(
+                "nix run {}#<name>",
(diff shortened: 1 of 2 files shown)


niksingh710:
Currently, the validity of a process is determined only by checking whether it contains a `package` and `settings` attribute.

Issue #416, opened 2025-02-10
`om ci`: Smarter extra args handling


srid:
This,

```
om ci run . -- --accept-flake-config
```

when run on Omnix, leads to:

<img width="722" alt="Image" src="https://github.com/user-attachments/assets/8cfb979d-a564-444a-a2df-278f47968c8a" />

The problem is that `extra_nix_build_args` is utilized only in `nix build` but not other commands like `nix flake lock` in the screenshot above.

https://github.com/juspay/omnix/blob/7da987ed6005da6e284d739b9672b81f1079edf7/crates/omnix-ci/src/step/build.rs#L107-L113

We need a better type to represent these arguments, and pass them intelligently to various nix commands. This also came up here: https://github.com/juspay/omnix/pull/399#discussion_r1926034162

srid:
These 3 options should also be consolidated in this new arguments handling:

https://github.com/juspay/omnix/blob/187547cd841cb0f0f0cd4f84c17efa40c4a98916/crates/nix_rs/src/command.rs#L35-L45

<img width="551" alt="Image" src="https://github.com/user-attachments/assets/d07d4b72-62e9-4773-b4df-5249c3f0b1d7" />

PR #417, opened 2025-02-10
Refactor arguments handling for `nix`

srid:
For #416 

`NixCmd` now delegates to a `NixArgs` struct that encapsulates all argument passing to `nix`. This then gets used in `RunCommand` clap subcommand struct, such that `om ci run` (and only that) inherits these options, per the screenshot below.

Consequently, the following now works:

```
om ci run . -- --accept-flake-config
```

because now `--accept-flake-config` is passed in **all** invocations of `nix`. Eventually, in next PRs, we will do it for other subcommands as well (`develop`, `health`, etc.).

image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,11 +53,10 @@ jobs:
       # Run omnix using self.
       - name: Omnix CI
         run: |
-          nix run . -- ci \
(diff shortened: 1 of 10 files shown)


srid:
I considered using `dyn` traits, but that would complicate the implementation.

Issue #204, opened 2024-08-06
`1.0` & nixpkgs release


srid:
## Before release 

See sub-issues below for pending work.

## Upstream

- [x] @srid Release to crates.io #423
- [x] Upstream `pkgs.omnix` to nixpkgs https://github.com/NixOS/nixpkgs/pull/385761
    - [x] Check: it installs the `om` binary (not `omnix-cli`).
    - [x] Check: auto completion works on zsh (macOS) and bash (Linux)
    - [x] Check: it builds on Intel mac 
      - Even if upstream doesn't test x86_64-darwin in CI, we should manually make sure that our PR builds.
    - [x] Check: `om ci run --on ssh://...` works.

srid:
All crates have now been published:

https://crates.io/crates/omnix-cli

srid:
Now that we no longer depend on the Nix fork, I'll work on preparing for release:

<img width="274" alt="image" src="https://github.com/user-attachments/assets/48e86933-8949-4643-ba6c-f2887292b397">

srid:
Regarding release version of `omnixrc`:

The caller can now override the omnix invocation as of https://github.com/juspay/omnix/commit/d25eab3a63f600e7c9db02e80b2a7c67c6a76f3a

Once we release 1.0 to nixpkgs, we can host the stable `omnixrc` at `https://ominx.page/omnixrc/1.0` (or some such versioned URL), which file would use `nix run nixpkgs/<commithash>#omnix` to fetch the pinned omnix.

When we release 1.1, we repeat the same process, allowing existings projects on 1.0 to remain unimpacted until they explicitly upgrade to 1.0 (by changing the URL and hash in their `.envrc`).

shivaraj-bh:
https://github.com/NixOS/nixpkgs/pull/385761

requires https://github.com/juspay/omnix/pull/427

srid:
Is there something we can do in the omnix source tree to avoid having to duplicate this in nixpkgs?

Perhaps a `.json` file that refers to these inputs, which the Nix expression can use directly.

<img width="598" alt="Image" src="https://github.com/user-attachments/assets/74c2b927-049f-4442-b981-4fdc74a99fa0" />

srid:
This is interesting.  I wonder if we can revert the [`doCheck = false;`](https://github.com/search?q=repo%3Ajuspay%2Fomnix%20doCheck&type=code) with this ...

[...]

PR #430, opened 2025-03-04
chore: Define ENVs in a single nix file

shivaraj-bh:
Avoids the need to explicitly define—and later maintain—these ENVs in nixpkgs while packaging for #204.

requires https://github.com/juspay/rust-flake/pull/33

### DIFF
--- a/nix/envs/default.nix
+++ b/nix/envs/default.nix
@@ -1,5 +1,5 @@
 { src, lib, cachix, fetchFromGitHub }:
-{
+lib.mapAttrs (_: v: builtins.toString v) {
   OMNIX_SOURCE = src;
   CACHIX_BIN = lib.getExe cachix;
   OM_INIT_REGISTRY = lib.cleanSourceWith {
--- a/nix/envs/default.nix
+++ b/nix/envs/default.nix
@@ -1,0 +1,75 @@
+{ src, lib, cachix, fetchFromGitHub }:
+{
+  OMNIX_SOURCE = src;
+  CACHIX_BIN = lib.getExe cachix;
+  OM_INIT_REGISTRY = lib.cleanSourceWith {
+    name = "om-init-registry";
+    src = src + /crates/omnix-init/registry;
+  };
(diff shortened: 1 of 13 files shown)


shivaraj-bh:
image

Ah, this doesn’t really work in `nixpkgs`. Fails with:

image

srid:
Convert it to path string or use `builtins.toString`

nix/modules/flake/rust.nix
```diff
@@ -40,8 +41,11 @@
             || lib.hasSuffix "flake-schemas/flake.lock" path
             || lib.hasSuffix "addstringcontext/flake.nix" path
             || lib.hasSuffix "addstringcontext/flake.lock" path
+            || lib.hasSuffix "metadata/flake.nix" path
+            || lib.hasSuffix "metadata/flake.lock" path
           ;
         };
+      defaultCraneArgs = import "${inputs.self}/nix/envs" { inherit (config.rust-project) src; inherit (pkgs) cachix fetchFromGitHub lib; };
```
srid, line 48: Better to follow `haskell-flake` convention:

```suggestion
      defaults.crate.crane.args = import "${inputs.self}/nix/envs" { inherit (config.rust-project) src; inherit (pkgs) cachix fetchFromGitHub lib; };
```

nix/envs/default.nix
```diff
@@ -0,0 +1,75 @@
+{ src, lib, cachix, fetchFromGitHub }:
+{
```
srid, line 2: This may fix the nixpkgs issue:

```suggestion
lib.mapAttrs (_: v: builtins.toString v) {
```

shivaraj-bh: works in nixpkgs

nix/modules/flake/rust.nix
shivaraj-bh: https://github.com/juspay/omnix/pull/430/commits/344f6dca79ef335f21860e0125778209fe3750e9

srid (approved):

PR #431, opened 2025-03-05
chore: Fix broken `watch_file` paths in `.envrc`

shivaraj-bh:


srid (changes requested):
This PR also changes the reload behaviour. We only want these 4 flake-parts module to trigger reload, not others (like `closure-size.nix`). I should have commented about it in the file.

image

Issue #397, opened 2025-01-21
CI: Avoid running impure tests


srid:
It is just waste of time due to double compilation. Can we disable sandox so these tests can be run in pure Nix?

https://github.com/juspay/omnix/blob/1bcb894ddc4b8021bc0dfbc3894f1480c992a58c/om.yaml#L20-L24

PR #432, opened 2025-03-05
chore: Enable Nix sandbox tests where applicable

shivaraj-bh:


crates/omnix-cli/crate.nix
```diff
@@ -32,9 +33,28 @@ in
...
+        # error: creating directory '/nix/var/nix/profiles': Permission denied
+        # TODO: elaborate on the error
+        "--skip=command::init::om_init"
+
+        # FIXME: This should pass in the sandbox
+        # > ❄\u{fe0f}  nix -j auto eval --json --override-input flake ../.. --override-input flake-schemas /nix/store/0npxvlidr0w1b99lrvcmd2wd46y2kibh-flake-schemas --override-input systems /nix/store/g1pv78p6lk92hjzr7syrcihvj4rx1fnz-source --no-write-lock-file \'/nix/store/qm8q4kylp1jy68k34c2bskpamxgyc9ix-source#contents.excludingOutputPaths\' --quiet --quiet\u{fe0f}
+        # > error: experimental Nix feature \'nix-command\' is disabled; use \'--extra-experimental-features nix-command\' to override
+        # > Error: Unable to fetch flake
+        "--skip=command::show::om_show_local"
+      ]);
```
shivaraj-bh, line 57: @srid This could be a potential bug. I even tried looking at how `om ci` works, but even that uses the same `NixCmd::Default`, so it is likely that even `om ci` doesn’t automatically apply the `experimental-features` when not enabled by default.

We don’t have a similar `om_ci_local` test (that can run in Nix sandbox), which could be why it went undetected.

shivaraj-bh:
Moved clippy changes to https://github.com/juspay/omnix/pull/433

srid:
Note: once #433 is merged, we may be able to replace the `om.yaml` run with [`rust-flake` clippy check](https://github.com/search?q=repo%3Ajuspay%2Frust-flake%20clippy&type=code) (if all the crates are being autowired).

srid (commented):
On second thought, I'm not sure if we should even bother to do this if *several* tests are being disabled anyway (for e.g., due to the fact that some "requires networking").

srid:
See also #397

We should run or the other, not both. If we can't run the full suite of tests in Nix, let's run tests outside of Nix only.

PR #453, opened 2025-05-20
chore(cache): Change trusted-public-key

shivaraj-bh:
The cache state was destroyed along with the attic server

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,7 @@
 {
   nixConfig = {
     extra-substituters = "https://cache.nixos.asia/oss";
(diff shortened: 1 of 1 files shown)


shivaraj-bh:
Cache info obtained using:
```sh
nix run nixpkgs#attic-client cache info oss
               Public: true
           Public Key: oss:KO872wNJkCDgmGN3xy9dT89WAhvv13EiKncTtHDItVU=
Binary Cache Endpoint: https://cache.nixos.asia/oss
         API Endpoint: https://cache.nixos.asia/
      Store Directory: /nix/store
             Priority: 41
  Upstream Cache Keys: ["cache.nixos.org-1"]
     Retention Period: Global Default
```

Issue #455, opened 2025-06-17
Nix flakes are not enabled for determinate installer


Aerma7309:
When using determinate installer and checking `om health` it says `Nix flakes are not enabled`

<img width="1563" alt="Image" src="https://github.com/user-attachments/assets/131acc16-6216-4726-9770-e8fe7d3d413b" />

PR #456, opened 2025-06-17
#455 fix : nix flake not enabled for determinate installer

Aerma7309:
Nix `flakes` and `nix-command` is enabled by default for determinate nix installations. Do not check them in `experimental-features` list when determinate install is present.

image

### DIFF
--- a/crates/omnix-health/src/check/flake_enabled.rs
+++ b/crates/omnix-health/src/check/flake_enabled.rs
@@ -18,8 +18,12 @@ impl Checkable for FlakeEnabled {
         let check = Check {
             title: "Flakes Enabled".to_string(),
             info: format!("experimental-features = {}", val.join(" ")),
-            result: if val.contains(&"flakes".to_string())
-                && val.contains(&"nix-command".to_string())
+            // determinate installer has flakes enabled by default
+            result: if matches!(
+                nix_info.nix_env.installer,
+                nix_rs::env::NixInstaller::DetSys(_)
+            ) || (val.contains(&"flakes".to_string())
+                && val.contains(&"nix-command".to_string()))
             {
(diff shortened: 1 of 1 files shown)


srid (commented):
@Aerma7309 How do we reproduce the original issue you are attempting to fix with this PR?

I used DetSys installer a few days ago (using https://nixos.asia/en/install) and never encountered this. My `experimental-features` is indeed full:

image

shivaraj-bh:
Judging by the nix version, `3.6.5`, looks like it is determinate nix

shivaraj-bh:
> > looks like it is determinate nix
> 
> That's also what I have (albeit `3.6.2` -- see screenshot above) yet I do not face OP's problem.

I should've elaborated. I meant the `nix` CLI version under the "Nix Version is supported" heading

Aerma7309:
#457 updated steps to reproduce

srid:
> #457 updated steps to reproduce

I can reproduce this on a macOS VM. So they moved it to `extra-experimental-features`? Does anyone have context behind this change?

I want to understand why before accepting any workaround or fix to address it.

Aerma7309:
can you suggest me a way to get the value of extra-experimental-features using config, 

also I found that even if i add the `nix-command flakes`  in  `/etc/nix/nix.custom.conf` as `experimental-features = auto-allocate-uids nix-command flakes`  

`nix config show | grep experimental` gives `experimental-features = auto-allocate-uids` this puzzles me more. 


image

srid (changes requested):
@Aerma7309 Can you update this PR to also look at `extra-experimental-features` (in addition to `experimental-features`)? Then you don't need to current change special casing on DetSys installer.

PR #460, opened 2025-06-23
fix: replaced getExe with getExe' to avoid nix warning.

niksingh710:
image

This will avoid this warning.


shivaraj-bh (approved):
diff LGTM

Issue #472, opened 2025-08-11
`om health` shows false negative against DetSys graphical installer


srid:
https://app.warp.dev/session/f4ecb03a-3f27-4579-bedd-95e3e153c7e9?pwd=e0b9dc31-ee0e-45e7-9ae6-91bfea005d5a

niksingh710:
When you say **DetSys graphical installer**, are you referring to the `.pkg` file at [pkgfile](https://install.determinate.systems/determinate-pkg/stable/Universal), the same one mentioned in the [blog post](https://determinate.systems/blog/graphical-nix-installer/)?  

I used the linked installer to install Nix on a clean macOS Silicon virtual machine and got the results shown below:  

| `nixone health curl` | `om health` |
|----------------------|-------------|
| <img width="1194" height="774" alt="Image" src="https://github.com/user-attachments/assets/48705400-c3fe-4f9c-ba35-7d6da6d06087" /> | <img width="856" height="492" alt="Image" src="https://github.com/user-attachments/assets/8da655fc-2945-4891-9664-0c5bc261dba4" /> |

In **DetSys** Nix, `flakes` are no longer considered *experimental*, which is why they do not appear in the output of:  
```bash
nix config show | grep experimental
````

For **`om`**, which approach should I take?

1. Run:

   ```bash
   nix flake metadata nixpkgs
   ```

   and check the exit code to determine if flakes are enabled.

2. Check if the Nix version/installation source indicates it’s from DetSys, and if so, automatically return flake check as `true` with a message saying `For DetSys flakes are not experimental`.

| Normal Nix | Normal Nix (by DetSys Installer) | DetSys Nix |
|-|-|-|
| `nix (Nix) 2.29.1` | `nix (Nix) 2.28.4` | `nix (Determinate Nix 3.8.5) 2.30.2` |

Output of diff `nix --version`

PR #473, opened 2025-08-12
feature: Flake check for detSys installed nix

niksingh710:
Resolves: #472 (based on the approach `2` mentioned in #472)

## On MacOs virtual Machine
image

## On Ubuntu virtual Machine (Simple Nix installation)
image

## Commands used to test the package for backward compatibility.

- Run Version Parsing Tests: `cargo test --package nix_rs test_parse_nix_version`
- Run Flake Detection Tests: `cargo test --package omnix-health flake_enabled`
- Test Core Packages `cargo test --package nix_rs --package omnix-health`
- Build and Test Health Command `cargo build --bin om`
- Verify JSON Output `./target/debug/om health --json`

## Backward Compatibility
✅ No breaking changes - all existing functionality preserved
✅ Regular Nix installations continue to work as before
✅ Adds new capability without removing existing features

## Expected Behavior

- **DetSys Nix users**: Health check shows "✅ Flakes Enabled" with message "Flakes enabled via Determinate Systems Nix"
- **Regular Nix users**: Existing behavior unchanged - checks `experimental-features` config

### DIFF
--- a/crates/nix_rs/src/version.rs
+++ b/crates/nix_rs/src/version.rs
@@ -9,6 +9,15 @@ use tracing::instrument;
 
 use crate::command::{NixCmd, NixCmdError};
 
+/// Nix installation type
+#[derive(Clone, Copy, PartialOrd, PartialEq, Eq, Debug)]
+pub enum NixInstallationType {
+    /// Official Nix installation
+    Official,
+    /// Determinate Systems Nix installation
+    DeterminateSystems,
+}
+
 /// Nix version as parsed from `nix --version`
 #[derive(Clone, Copy, PartialOrd, PartialEq, Eq, Debug, SerializeDisplay, DeserializeFromStr)]
 pub struct NixVersion {
@@ -18,8 +27,8 @@ pub struct NixVersion {
     pub minor: u32,
     /// Patch version
     pub patch: u32,
-    /// Whether this is a Determinate Systems Nix installation
-    pub is_detsys: bool,
+    /// Nix installation type
+    pub installation_type: NixInstallationType,
 }
 
 /// Error type for parsing `nix --version`
@@ -52,14 +61,18 @@ impl FromStr for NixVersion {
         let minor = captures[2].parse::<u32>()?;
         let patch = captures[3].parse::<u32>()?;
 
-        // Check if this is a Determinate Systems Nix installation
-        let is_detsys = s.contains("Determinate");
+        // Determine the installation type based on the version string
+        let installation_type = if s.contains("Determinate") {
+            NixInstallationType::DeterminateSystems
+        } else {
+            NixInstallationType::Official
+        };
 
         Ok(NixVersion {
             major,
             minor,
             patch,
-            is_detsys,
+            installation_type,
         })
     }
 }
@@ -108,7 +121,7 @@ async fn test_parse_nix_version() {
             major: 2,
             minor: 13,
             patch: 0,
-            is_detsys: false
+            installation_type: NixInstallationType::Official
         })
     );
 
@@ -119,7 +132,7 @@ async fn test_parse_nix_version() {
             major: 2,
             minor: 13,
             patch: 0,
-            is_detsys: false
+            installation_type: NixInstallationType::Official
         })
     );
 
@@ -130,7 +143,7 @@ async fn test_parse_nix_version() {
             major: 2,
             minor: 29,
             patch: 0,
-            is_detsys: true
+            installation_type: NixInstallationType::DeterminateSystems
         })
     );
 }
--- a/crates/nix_rs/src/version.rs
+++ b/crates/nix_rs/src/version.rs
@@ -54,26 +54,37 @@ impl FromStr for NixVersion {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         // NOTE: The parser is lenient in allowing pure nix version (produced
         // by [Display] instance), so as to work with serde_with instances.
-        let re = Regex::new(r"(?:nix \(.*?\) )?(\d+)\.(\d+)\.(\d+)$")?;
 
-        let captures = re.captures(s).ok_or(BadNixVersion::Command)?;
-        let major = captures[1].parse::<u32>()?;
-        let minor = captures[2].parse::<u32>()?;
-        let patch = captures[3].parse::<u32>()?;
+        // Define patterns and their corresponding installation types
+        let patterns = [
+            (
+                r"^nix \(Determinate Nix [\d.]+\) (\d+)\.(\d+)\.(\d+)$",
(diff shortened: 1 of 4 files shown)


crates/nix_rs/src/version.rs
```diff
@@ -43,17 +45,21 @@ impl FromStr for NixVersion {
...
-        let re = Regex::new(r"(?:nix \(Nix\) )?(\d+)\.(\d+)\.(\d+)$")?;
+        let re = Regex::new(r"(?:nix \(.*?\) )?(\d+)\.(\d+)\.(\d+)$")?;
 
         let captures = re.captures(s).ok_or(BadNixVersion::Command)?;
         let major = captures[1].parse::<u32>()?;
         let minor = captures[2].parse::<u32>()?;
         let patch = captures[3].parse::<u32>()?;
 
+        // Check if this is a Determinate Systems Nix installation
+        let is_detsys = s.contains("Determinate");
```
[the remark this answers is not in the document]
srid: Strings are bad. Use a sum type.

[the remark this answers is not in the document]
srid: https://en.wikipedia.org/wiki/Algebraic_data_type

crates/nix_rs/src/version.rs
```diff
@@ -43,17 +54,25 @@ impl FromStr for NixVersion {
...
         let minor = captures[2].parse::<u32>()?;
         let patch = captures[3].parse::<u32>()?;
 
+        // Determine the installation type based on the version string
+        let installation_type = if s.contains("Determinate") {
+            NixInstallationType::DeterminateSystems
+        } else {
+            NixInstallationType::Official
+        };
+
```
srid, line 70: Needs to be more rigorous than this. And fail for any unexpected string.

crates/nix_rs/src/version.rs
```diff
@@ -18,6 +27,8 @@ pub struct NixVersion {
     pub minor: u32,
     /// Patch version
     pub patch: u32,
+    /// Nix installation type
+    pub installation_type: NixInstallationType,
 }
```
srid, line 32: What effect does this have on comparing two `NixVersion` with same version but different installation type?

crates/omnix-health/src/check/flake_enabled.rs
```diff
@@ -15,22 +15,156 @@ impl Checkable for FlakeEnabled {
         _: Option<&nix_rs::flake::url::FlakeUrl>,
     ) -> Vec<(&'static str, Check)> {
         let val = &nix_info.nix_config.experimental_features.value;
+
+        // Check if flakes are enabled either through config or DetSys installation
+        let flakes_enabled = match nix_info.nix_version.installation_type {
+            NixInstallationType::DeterminateSystems => {
+                // Determinate Systems Nix has flakes enabled by default
+                true
```
srid, line 23: Is it possible to disable flakes on DetSys installer? And if so, does the PR behave correctly?

crates/omnix-health/src/check/flake_enabled.rs
```diff
@@ -15,22 +15,156 @@ impl Checkable for FlakeEnabled {
...
+    fn mock_nix_config_without_flakes() -> NixConfig {
+        NixConfig {
+            cores: mock_config_val(4),
+            experimental_features: mock_config_val(vec!["auto-allocate-uids".to_string()]),
+            extra_platforms: mock_config_val(vec![]),
+            flake_registry: mock_config_val("https://flake-registry.example.com".to_string()),
+            max_jobs: mock_config_val(8),
+            substituters: mock_config_val(vec![Url::parse("https://cache.nixos.org/").unwrap()]),
+            system: mock_config_val(System::Linux(Arch::X86_64)),
+            trusted_users: mock_config_val(vec![TrustedUserValue::User("root".to_string())]),
```
srid, line 94: Why do you need all these irrelevant settings in the mock config? KISS and as small a code as possible.

crates/nix_rs/src/version.rs
niksingh710: For example:
```rust
- NixVersion { major: 2, minor: 13, patch: 0, installation_type: Official }
- NixVersion { major: 2, minor: 13, patch: 0, installation_type: DeterminateSystems }
```
The above two are no longer equal because their `installation_type` values differ.

For ordering (`>`, `>=`, `<`, `<=`):
- Rust’s derived `PartialOrd` compares fields in declaration order.
- It compares `major`, then `minor`, then `patch`, and finally `installation_type`.
- Since `Official` is declared before `DeterminateSystems` in the enum, `Official < DeterminateSystems`.

I believe this behavior is fine, which is why I haven’t implemented custom `PartialOrd` or `Eq` for `NixVersion` comparisons. If you think it’s necessary, I can implement them as well.

niksingh710: I tested this on both a virtual device and my personal system with different versions in the `yaml` file, and confirmed that the `nix_version` check works correctly, since Rust’s `PartialOrd` handles it well.

srid: How exhaustive was your check? Can you read the code around:

https://github.com/juspay/omnix/blob/99ddd5a50f892986a822fe2645faa6c7c5abc002/crates/nix_rs/src/version_spec.rs#L75

And see how the `>` check can fail or pass depending on the `installation_type` field?

The presence of `installation_type` seems quite problematic to me in this regard. `NixVersion` means just that ... nix **version** (conceptually, it says nothing about the provider of that installation).

crates/omnix-health/src/check/flake_enabled.rs
```diff
@@ -15,22 +15,114 @@ impl Checkable for FlakeEnabled {
...
+            r#"{{
+            "experimental-features": {{ "value": [{}], "defaultValue": [], "description": "" }},
+            "cores": {{ "value": 1, "defaultValue": 1, "description": "" }},
+            "extra-platforms": {{ "value": [], "defaultValue": [], "description": "" }},
+            "flake-registry": {{ "value": "", "defaultValue": "", "description": "" }},
+            "max-jobs": {{ "value": 1, "defaultValue": 1, "description": "" }},
+            "substituters": {{ "value": [], "defaultValue": [], "description": "" }},
+            "system": {{ "value": "x86_64-linux", "defaultValue": "x86_64-linux", "description": "" }},
+            "trusted-users": {{ "value": [], "defaultValue": [], "description": "" }}
+        }}"#,
```
srid, line 74: [Just](https://github.com/juspay/omnix/pull/473#discussion_r2283433150) mock the `nix.conf` file itself (keep it 1 line; for experimental-features), and then parse it.

`nix.conf` is in INI format, not JSON, as you are mocking here.

Issue #474, opened 2025-08-13
`om develop`: Support `attic`


shivaraj-bh:
`attic use` should be run in the `preShell` if an attic cache is configured in `om.health.caches.required`, like the existing behaviour when a `cachix` cache is configured. 

There is a challenge here: We cannot auto-detect `attic` caches like we do for `cachix` from the domain name (`<name>.cachix.org`).

srid:
Sounds good, let's do it.

PR #475, opened 2025-08-13
feat: Add attic support in om develop

niksingh710:
Resolves: #474


srid (commented):
Just took a glance. AI generated code is almost always slop on first pass. Have you refactored it already?

shivaraj-bh (changes requested):

